PR #9, opened 2024-11-08
Introduce `BlobIO` interface

kyleknap:
This will be the primary file-like object interface used for integration with PyTorch. This allows customers to have direct access to a file-like interface to read or write blobs from Azure Storage. These can be directly used as part of checkpointing (e.g. `torch.load` and `torch.save`) and load training data.

Below are important notes to keep in mind when reviewing the pull request:
* The pull request focuses on establishing the interface for the file-like object. It currently uses a stubbed downloader class to illustrate how the interface may integrate with SDK usage when reading. A follow up pull request will be made to actually flesh out the downloader implementation and thus this integration may change based on findings.
* The pull request does not attempt to fully implement the interface for writes. The interface and implementation for writes will be introduced at a later time.
* Most of the behavior was derived from following Python's documentation on [`IOBase`](https://docs.python.org/3/library/io.html#io.IOBase) and comparing with existing standard-libary implementations. As a sanity check, I took the test that I built out and swapped it with a [`BytesIO`](https://github.com/kyleknap/azure-storage-for-pytorch/commit/6346a3d911091cbbd61423a3f7692910d4c7e67b) and [file-like from `open()`](https://github.com/kyleknap/azure-storage-for-pytorch/commit/4ee9424425a443f52bbec96e034d473cd3dc9e3c) to see how well the interface matches. Overall, it matched fairly well with mainly error message differences and read sizes less than `-1` being supported for `BytesIO` but not `open()` (decided to stick with open()'s behavior for `BlobIO`)
* There are some additional improvements that I'd like to eventually add such as docstrings, type checking, formatters, end-to-end tests. These were deferred to keep this initial pull request lean, allowing us to focus on just the interface.
* For the generated `requirements-dev.txt` lock file, I unfortunately already ran into complications in being able to keep a single requirements file and solely generate it using `pip-compile`. Specifically, in pulling in `azure-identity` there are windows-only dependencies that are not valid on Linux. To handle this, I just manually added the windows environment marker for now to get CI passing. I made this a separate commit as well to make it easier to see what I changed. We should revisit this by either creating two requirements files (one for windows and one for linux) or building a script that programatically adds the marker after running `pip-compile`.


src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def read(self, size: Optional[int] = -1, /) -> bytes:
+        if size is not None:
+            self._validate_is_integer("size", size)
+            self._validate_min("size", size, -1)
+        self._validate_not_closed()
+        return self._read(size)
+
+    def readable(self) -> bool:
+        self._validate_not_closed()
```
jalauzon-msft, line 68: Is it typical for `readable` to raise if the IO is closed? I feel like it should always just return whether this type is readable or not but then again, I've never tried this on a standard IO.

src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+        self._validate_is_integer("offset", offset)
+        self._validate_is_integer("whence", whence)
+        self._validate_not_closed()
+        return self._seek(offset, whence)
+
+    def seekable(self) -> bool:
+        return self.readable()
+
+    def tell(self) -> int:
+        self._validate_not_closed()
```
jalauzon-msft, line 93: Same question here, is it typical for this to raise when the stream is closed?

src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+    def closed(self) -> bool:
+        return self._closed
+
+    def fileno(self) -> int:
+        raise OSError("BlobIO object has no fileno")
+
+    def flush(self) -> None:
+        self._validate_not_closed()
+
+    def read(self, size: Optional[int] = -1, /) -> bytes:
```
jalauzon-msft, line 60: Does size need to be `Optional` here? Would `None` be different than `-1`?

Or is my understanding of `Optional` not right. I always thought it just expands to `Union[<type>, None]` in which case `int` would be sufficient here.

src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+            return False
+        parsed_qs = urllib.parse.parse_qs(parsed_url.query)
+        # The signature is always required in a valid SAS token. So look for the "sig"
+        # key to determine if the URL has a SAS token.
+        return "sig" in parsed_qs
+
+    def _get_blob_downloader(self) -> "_BlobDownloader":
+        return _BlobDownloader(self._blob_url, self._sdk_credential)
+
+    def _read(self, size: Optional[int]) -> bytes:
```
jalauzon-msft, line 143: If we take away the `Optional` in the public `read` we may not need it here either?

src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+        if size == 0 or self._position >= self._blob_downloader.get_blob_size():
+            return b""
+        download_length = size
+        if size is not None and size < 0:
+            download_length = None
```
jalauzon-msft, line 148: I assume `download_length=None` will mean all?

src/azstoragetorch/io.py
kyleknap: Before sending out the PR, I did a pass on how closely `BlobIO` behaves to other standard library file-like objects. There was definitely some unexpected behavior and this was one of them. Specifically, whether it throws an exception is actually based on the mode that is open. Below shows some things I tried out:
```python
# Usage on writable file
>>> with open('hello.txt', 'wb') as f:
...     pass
>>> f.writable()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file
>>> f.readable()
False

# Example usage on readable
>>> with open('hello.txt', 'rb') as f:
...     pass
>>> f.writable()
False
>>> f.readable()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file

# Example usage on something that is readable and writable
>>> import io
>>> f = io.BytesIO(b'foo')
>>> f.close()
>>> f.readable()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file.
>>> f.writable()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file.
```
Specifically, it appears that it only throws an error if the file was closed and the `readable()`/`writable()` would have returned `True`. Thinking about it more, it seems like the standard-library is taking the perspective that when calling these methods, it is with the intention that the user will call one of the followup methods (e.g. `read()` or `write()`) and thus raise the error sooner to avoid a bad state/unintended logic being triggered because the file is closed.

I think it makes sense to keep parity with this behavior. Though the way it is implemented in `readable()` will not be correct once `wb` is implemented. So happy future-proof the `readable()` method as part of this PR to avoid forgetting to update it.

src/azstoragetorch/io.py
kyleknap: Yeah same thing here as in `readable()`; it's what the standard library file-like objects do:
```python
>>> import io
>>> f = io.BytesIO(b'foo')
>>> f.close()
>>> f.tell()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file.
>>> with open('hello.txt', 'rb') as f:
...     pass
...
>>> f.tell()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: I/O operation on closed file
```
I think this behavior makes sense because if the file is closed, there is not really a current position in the file. So raising an error is better than returning a position that is not technically correct.

src/azstoragetorch/io.py
kyleknap: Yeah optional is still needed here. Technically, someone can pass in `None`:
```
blob_io.read(None)
```
and explicitly `-1`:
```
blob_io.read(-1)
```
If the `Optional` was not there, it would fail mypy checking if someone passed in `None` because `None` is not captured by `int` type. It also how typeshed types some of the `io` implementation as well. For example, this is the typing for BufferedIOBase: https://github.com/python/typeshed/blob/af4df4eb24f1cda42033f74440ba1117ce4c8f7e/stdlib/_io.pyi#L62

jalauzon-msft (approved):

vincenttran-msft (approved):
LGTM 😄 Just one pending question regarding blob types

src/azstoragetorch/io.py
kyleknap: Talked offline about this. Yep. It means read all. I was trying to match the values used in the SDK for this downloader class.

src/azstoragetorch/io.py
kyleknap: I pushed up an update to `readable()` to make this a little more futureproofed as I suggested when we add support for `wb`.

src/azstoragetorch/io.py
```diff
@@ -0,0 +1,188 @@
...
+_SDK_CREDENTIAL_TYPE = Optional[
+    Union[
+        AzureSasCredential,
+        TokenCredential,
+    ]
+]
+_AZSTORAGETORCH_CREDENTIAL_TYPE = Union[_SDK_CREDENTIAL_TYPE, Literal[False]]
+
+
+class BlobIO(io.IOBase):
```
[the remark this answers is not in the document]
kyleknap: > Is the concept of different blob types relevant here? 

Possibly. If we did want to support different blob types for writes (for the initial implementation, it was going to be block blob support only), I'd first suggest exploring accepting the blob_type as part of the `BlobIO` constructor e.g.:
```python
BlobIO('<blob-url>', 'wb', blob_type='page')
```
The `BlobIO` would have the same file-like object interface methods. But under the hood, we could delegate to different uploader implementations based on what type of blob was specified and we could keep the write logic needed for each blob type to their own uploader class. It would be sort of similar approach to how the Python SDK has different class implementations through a [common chunk uploader interface](https://github.com/Azure/azure-sdk-for-python/blob/7b18366f341b2b85ee10397387315a3e3bec40c4/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py#L120).

There is probably some other interfaces we can explore here such as having explicit `BlobIO` classes based on the blob type (e.g. `PageBlobIO`) or even exploring if it makes sense to map the `mode` to a blob type (e.g. `ab` would make it an append blob). Nonetheless, I think there are ways we should be able to keep the file-like methods the same as the standard library but still be able to specify a blob type if we decide we need it.

Thoughts?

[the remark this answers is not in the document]
vincenttran-msft: Gotcha, this all makes sense and should not stop any further developments, as this initial interface doesn't interfere with introducing blob types later on.

I agree that a lightweight initial implementation with just block blob support is a good start so we can get deep on getting it working fully E2E, then we can revisit and fold in concepts like blob types and adjust accordingly.

PR #10, opened 2025-02-11
Implement `read()` for BlobIO

kyleknap:
Implementation includes:
* Refactoring of BlobDownloader interface to azstoragetorch's own blob client class. This will be useful when we implement as we will not have separate classes for upload/download classes which makes it easier for dependency injection and keeping all Azure Blob specific logic to a single class.
* Relying on code generated clients for downloads. This was done to utilize a leaner download implementation for notable performance improvements over the public `download_blob()` method.

Note this pull request does not implement `readline()`. That will be done in a follow up PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "azure-identity<2",
+    "azure-storage-blob<13"
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,11 @@ classifiers = [
 ]
 dependencies = [
     "azure-identity<2",
-    "azure-storage-blob<13"
+    # This is pinned because the library uses internal APIs of the blob SDK and helps
+    # avoid automatically breaking the library if a new version of the SDK breaks the
+    # internal API. As new versions of the SDK are released, the ceiling of this version
+    # range should be updated to match the new version.
+    "azure-storage-blob>=12.24.0,<=12.24.1"
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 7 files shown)


src/azstoragetorch/_client.py
```diff
@@ -0,0 +1,146 @@
...
+            start = offset + i * self._PARTITION_SIZE
+            if start >= end:
+                break
+            size = min(self._PARTITION_SIZE, end - start)
+            partitions.append((start, size))
+        return partitions
+
+    def _download_with_retries(self, pos: int, length: int) -> bytes:
+        attempts_remaining = 3
+        while attempts_remaining > 0:
```
jalauzon-msft, line 113: We should probably add some sleep time in between tries. Retrying without sleep may not help and may exasxcerbate network issues.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ classifiers = [
 ]
 dependencies = [
     "azure-identity<2",
+    "azure-storage-blob<13"
```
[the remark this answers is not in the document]
kyleknap: This is a good callout. I put a version range with a ceiling on the latest version of the SDK that we can increment. Even though the APIs seem somewhat stable, there needs to be a safety net there in case it did change.

src/azstoragetorch/_client.py
kyleknap: I added exponential backoff with full jitter as a starting point here. Mainly it will at least pause and slightly stagger streaming across workers. In the future, if connection issues become an issue especially because of overwhelming the network, we should consider adapting how fast we read from the stream as an approach to reducing the network errors. I added this as a comment as well to help remind us if we revisit it.

jalauzon-msft (approved):

vincenttran-msft (approved):
:shipit:

PR #11, opened 2025-02-20
Implement `readline()` for BlobIO

kyleknap:
Includes prefetching and caching to avoid calling `download()` one byte at a time. For prefetching, a fixed amount is used to have similar behavior as the standard library io module, where prefetch is a separate configuration from size. In the future, we can make the prefetch amount configurable.

### DIFF
--- a/src/azstoragetorch/io.py
+++ b/src/azstoragetorch/io.py
@@ -24,6 +24,9 @@ _AZSTORAGETORCH_CREDENTIAL_TYPE = Union[_SDK_CREDENTIAL_TYPE, Literal[False]]
 
 
 class BlobIO(io.IOBase):
+    _READLINE_PREFETCH_SIZE = 4 * 1024 * 1024
+    _READLINE_TERMINATOR = b"\n"
+
     def __init__(
         self,
         blob_url: str,
@@ -45,6 +48,7 @@ class BlobIO(io.IOBase):
 
         self._position = 0
         self._closed = False
+        self._readline_buffer = b""
 
     def close(self) -> None:
         self._close_client()
@@ -65,6 +69,7 @@ class BlobIO(io.IOBase):
             self._validate_is_integer("size", size)
             self._validate_min("size", size, -1)
         self._validate_not_closed()
+        self._invalidate_readline_buffer()
         return self._read(size)
 
     def readable(self) -> bool:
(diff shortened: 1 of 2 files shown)


src/azstoragetorch/io.py
```diff
@@ -153,8 +158,58 @@ def _blob_url_has_sas_token(self, blob_url: str) -> bool:
...
+                offset=self._position, length=self._READLINE_PREFETCH_SIZE
+            )
+            consumed = self._consume_from_readline_buffer(consumed, limit)
+        return consumed
+
+    def _get_limit(self, size: Optional[int]) -> Optional[int]:
+        limit = size
+        if size is not None and size < 0:
+            limit = None
+        return limit
```
jalauzon-msft, line 180: Not sure how you'll feel about this but it may be easier to use `sys.maxsize` instead of `None` if there is no limit? It just means the return type can't be `None` which makes working with it a little easier later on.

src/azstoragetorch/io.py
```diff
@@ -153,8 +158,58 @@ def _blob_url_has_sas_token(self, blob_url: str) -> bool:
...
+        find_pos = self._readline_buffer.find(self._READLINE_TERMINATOR, 0, limit)
+        end = find_pos + 1
+        if find_pos == -1:
+            buffer_length = len(self._readline_buffer)
+            if limit is None:
+                end = buffer_length
+            else:
+                end = min(buffer_length, limit)
+        consumed += self._readline_buffer[:end]
+        self._readline_buffer = self._readline_buffer[end:]
```
jalauzon-msft, line 196: Not sure how much we care here but `_readline_buffer`is a `bytes` and so slicing it does a copy right? Would it be worth using a `bytearray` to avoid the copies?

jalauzon-msft (approved):

src/azstoragetorch/io.py
kyleknap: I definitely agree it would make the logic simpler by reducing branching. I'm hesitant to use `sys.maxsize` because while Blobs do not support blobs of that size, it does not mean it won't in the future. Instead, I'm thinking we could use the blob size instead as the limit, which is already available on the client and we grab and cache anyways on downloads.

src/azstoragetorch/io.py
kyleknap: I'm not sure if a `bytearray` will help much here mainly it is just a mutable form of `bytes`. It still returns a copy of the bytes when sliced in order to create a new `bytearray` object. I think exploring using [`memoryview`](https://docs.python.org/3/library/stdtypes.html#memory-views) objects could be interesting since those do not copy on splices. I'll add a note about considering converting it to a `memoryview`. I've been looking more into `memoryview`s as part of the write implementation so we can circle back to this later.

kyleknap: Talked more offline about this... The `bytearray` would help when we concatenate the readline buffer slice to `consumed` as there are performance benefits because the `consumed` bytearray would be mutable. That being said, we are not doing these computations in a tight loop; the download itself will take much more time. So any potential performance gain will be minimal if not negligible.

PR #12, opened 2025-03-05
Adding end to end read test

anjaliratnam-msft:
Allows us to test the read method against azure blob storage.


anjaliratnam-msft:
@microsoft-github-policy-service agree company="Microsoft"

tests/test_read.py
```diff
@@ -0,0 +1,70 @@
+# -------------------------------------------------------------------------
```
kyleknap, line 1: Currently CI is not passing because these tests require credentials and the environments are not configured with credentials. In order to get the CI passing, I'm thinking we do the following:

1. Create a new `unit` directory in the `tests` directory and add all of the files from the `tests` directory prior to this PR to this new `unit` subdirectory. This will represent all unit tests that we want to write for the project and not need credentials.
2. Create a new `e2e` directory in the `tests` directory and add this file to the new directory. This will represent all of the tests where we run end to end testing that hit live Azure storage.
3. Update this line so that pytests only run the tests in `tests/unit` to avoid running the end to end tests that we will not be running in CI for now.

tests/test_read.py
```diff
@@ -0,0 +1,70 @@
...
+from azstoragetorch._client import AzStorageTorchBlobClient
+
+from azure.core.exceptions import ResourceExistsError
+from azure.storage.blob import BlobServiceClient
+
+
+
+@pytest.fixture
+def account_url():
+    account_name = os.environ.get('PYTORCH_STORAGE_ACCOUNT_NAME', None) or "pytorchstorageaccount"
```
kyleknap, line 25: For this fixture, instead of setting a default account name, I'm think we just raise an error. So the fixture could look something like this:
```python
@pytest.fixture
def account_url():
    account_name = os.environ.get("AZSTORAGETORCH_STORAGE_ACCOUNT_NAME")
    if account_name is None:
        raise ValueError(
            f'"AZSTORAGETORCH_STORAGE_ACCOUNT_NAME" environment variable must be set to run end to end tests.'
        )
    return f"https://{account_name}.blob.core.windows.net"
```
The reason that I'm suggesting it is that if individuals run these tests, they likely do not have this environment variable set and also likely do not have access to the fall back storage account. So, it is both safer and clearer by explicitly raising an exception asking the operator to set the environment variable as otherwise they will just get authorization errors in not being able to access that account

kyleknap, line 25: For any environment variables we expose, let's stick with the prefix of `AZSTORAGETORCH_`. So for this one, it would be `AZSTORAGETORCH_STORAGE_ACCOUNT_NAME`. The main reason that I'm suggesting this is because we own the package name `azstoragetorch`, it makes it clear that this environment variable is specific to this library and not PyTorch the library.

tests/test_read.py
```diff
@@ -0,0 +1,70 @@
...
+    return "https://" + account_name.strip('"') + ".blob.core.windows.net"
+
+@pytest.fixture
+def container_name():
+    return ''.join(random.choices(string.ascii_lowercase + string.digits, k=8))
```
kyleknap, line 30: For this line, it may be worth introducing it as a helper function e.g. `random_resource_name()` so that we can call it in the different fixtures that need a randomized resource name and not have to individually repeat this logic for each fixture (and future fixture) that may need it.

tests/test_read.py
```diff
@@ -0,0 +1,70 @@
...
+    return ''.join(random.choices(string.ascii_lowercase + string.digits, k=8)).encode()
+
+@pytest.fixture
+def container(
+    account_url, container_name
+):
+    blob_service_client = BlobServiceClient(account_url, credential=DefaultAzureCredential())
+
+    cont = blob_service_client.create_container(name=container_name)
+    yield blob_service_client
```
kyleknap, line 47: For this fixture, lets:
1. Yield the container client back as part of the fixture. In general, the service client will likely not be needed for much past creating containers. So that will reduce the number of hops needed in downstream fixtures between client calls.
1. Rename it to `container_client`. This helps the fixture be a bit more self-documenting in that it is returning a container client as it could be interpreted that `container` could be a container name or url.

kyleknap (commented):
Nice! This looks good. Most of my feedback was minor and more just things to keep in mind especially for PRs in the future.

tests/e2e/test_read.py
```diff
@@ -0,0 +1,71 @@
...
+import random
+import string
+import pytest
+
+
+from azure.identity import DefaultAzureCredential
+
+from azstoragetorch.io import BlobIO
+
+from azure.storage.blob import BlobServiceClient
```
kyleknap, line 16: It's a little weird that `ruff` did not remove some of the extra newlines here. I'd recommend installing [`isort`](https://pycqa.github.io/isort/) and running it on this file to organize the imports more to the [pattern](https://peps.python.org/pep-0008/#imports) that is being followed in the repository.

tests/e2e/test_read.py
```diff
@@ -0,0 +1,71 @@
...
+    return f"{account_url}/{container_client.container_name}/{blob_client.blob_name}"
+
+
+@pytest.fixture
+def blob_io(blob_url):
+    yield BlobIO(
+        blob_url=blob_url, mode="rb"
+    )
+
+def random_resource_name(byte_num):
```
kyleknap, line 64: For the `random_resource_name()` function, we can probably rename the parameter to `name_length` and then make it's default `8` (e.g. `name_length=8`) so we do not need to specify the length every time we use it.

kyleknap (commented):
Looks good! I just had a few more comments, but after we should be good to merge it

kyleknap (approved):
Looks good! :ship:

PR #13, opened 2025-03-07
Adding end to end tests for reading n bytes, random seeks, __iter__, and large blobs

anjaliratnam-msft:
Tests the read function for reading n bytes from azure blob storage.

### DIFF
--- a/tests/e2e/test_read.py
+++ b/tests/e2e/test_read.py
@@ -111,20 +111,19 @@ class TestRead:
         with BlobIO(small_blob.url, "rb") as f:
             for i in range(0, len(small_blob.data), n):
                 assert f.read(n) == small_blob.data[i : i + n]
-                if n < len(small_blob.data):
-                    assert f.tell() == i + n
+                expected_position = min(i + n, len(small_blob.data))
+                assert f.tell() == expected_position
 
     @pytest.mark.parametrize("n", [1, 5, 20, 21])
     def test_random_seeks_and_reads(self, small_blob, n):
         with BlobIO(small_blob.url, "rb") as f:
             f.seek(n)
             assert f.read() == small_blob.data[n:]
-            if n < len(small_blob.data):
-                assert f.tell() == len(small_blob.data)
+            expected_position = max(n, len(small_blob.data))
+            assert f.tell() == expected_position
 
     def test_read_using_iter(self, small_with_newlines_blob):
         with BlobIO(small_with_newlines_blob.url, "rb") as f:
-            data = f.read().splitlines()
-            expected_data = small_with_newlines_blob.data.splitlines()
-            for i in data:
-                assert data == expected_data
+            lines = [line for line in f]
+            expected_lines = small_with_newlines_blob.data.splitlines(keepends=True)
+            assert lines == expected_lines
--- a/tests/e2e/test_read.py
+++ b/tests/e2e/test_read.py
@@ -45,46 +45,25 @@ def container_client(account_url):
 
 
 @pytest.fixture(scope="module")
-def small_uploaded_blob(account_url, container_client):
-    blob_name = random_resource_name()
-    data = sample_data(20)
-    blob_client = container_client.get_blob_client(blob=blob_name)
-    blob_client.upload_blob(data)
-    url = f"{account_url}/{container_client.container_name}/{blob_name}"
-
-    return Blob(data=data, url=url)
+def small_blob(account_url, container_client):
+    return upload_blob(account_url, container_client, sample_data(20))
 
 
 @pytest.fixture(scope="module")
-def large_uploaded_blob(account_url, container_client):
-    blob_name = random_resource_name()
-    data = sample_data(_PARTITIONED_DOWNLOAD_THRESHOLD * 2)
-    blob_client = container_client.get_blob_client(blob=blob_name)
-    blob_client.upload_blob(data)
-    url = f"{account_url}/{container_client.container_name}/{blob_name}"
-
-    return Blob(data=data, url=url)
+def large_blob(account_url, container_client):
+    return upload_blob(
+        account_url, container_client, sample_data(_PARTITIONED_DOWNLOAD_THRESHOLD * 2)
+    )
 
 
 @pytest.fixture(scope="module")
-def small_with_newlines_uploaded_blob(account_url, container_client):
-    blob_name = random_resource_name()
-    data = sample_data(20, True)
-    blob_client = container_client.get_blob_client(blob=blob_name)
-    blob_client.upload_blob(data)
-    url = f"{account_url}/{container_client.container_name}/{blob_name}"
-
-    return Blob(data=data, url=url)
+def small_with_newlines_blob(account_url, container_client):
+    return upload_blob(account_url, container_client, sample_data_with_newlines(20, 2))
 
 
 @pytest.fixture
-def uploaded_blob(request):
-    if request.param == "small_blob":
-        return request.getfixturevalue("small_uploaded_blob")
-    elif request.param == "large_blob":
-        return request.getfixturevalue("large_uploaded_blob")
-    elif request.param == "small_with_newlines_blob":
-        return request.getfixturevalue("small_with_newlines_uploaded_blob")
+def blob(request):
+    return request.getfixturevalue(f"{request.param}_blob")
 
 
 def random_resource_name(name_length=8):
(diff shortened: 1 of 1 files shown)


tests/e2e/test_read.py
```diff
@@ -58,10 +61,49 @@ def blob_io(blob_url):
...
+
+    @pytest.mark.parametrize(
+        "sample_data",
+        [
+            20,
+            1000,
+            _PARTITIONED_DOWNLOAD_THRESHOLD,
+            _PARTITIONED_DOWNLOAD_THRESHOLD * 2,
+        ],
+        indirect=True,
```
kyleknap, line 105: I definitely like the idea of using `indirect` here to utilize the fixtures as part of the setup. I think one concern that I have is that it means that we have to reupload the blob for each test case iteration, and it would be interesting to see if we can figure out how to reduce the number of times we need to reupload a blob and also reuse blobs with logical names across tests.

To do this, it would probably make sense to introduce `module` scoped blob fixtures for the different sizes that we want to test. For example, we can just have a handful of these (e.g. `small_uploaded_blob`, `large_uploaded_blob`, `small_with_newlines_uploaded_blob`) where the sample data is generated and uploaded once during this fixture setup.

As part of the return value for these, we could also introduce a `Blob` [dataclass](https://docs.python.org/3/library/dataclasses.html#module-contents) that will serve to hold properties useful for the test such as the `url` and the `expected_contents`. We can also add more useful properties to this in the future but it would mainly help group blob specific values together without needing to spread it out too much across fixtures.

Finally we could then add a top-level fixture `uploaded_blob` that calls [`request.getfixturevalue()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.FixtureRequest.getfixturevalue) that returns the uploaded blob based on the name.

So by the end of it this test case could look like:
```python
    @pytest.mark.parametrize(
        "uploaded_blob",
        [
            "small_blob",
            "large_blob",
        ],
        indirect=True,
    )
    def test_read_all_data(self, uploaded_blob):
        with BlobIO(uploaded_blob.url, "rb") as f:
            assert f.read() == uploaded_blob.expected_contents
```

kyleknap (commented):
This is looking good! My main comment was related to restructuring some of the fixtures that we have to make the cases more usable and also make the tests run faster.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
 @pytest.fixture
-def blob_io(blob_url):
-    yield BlobIO(blob_url=blob_url, mode="rb")
+def uploaded_blob(request):
+    if request.param == "small_blob":
+        return request.getfixturevalue("small_uploaded_blob")
+    elif request.param == "large_blob":
+        return request.getfixturevalue("large_uploaded_blob")
+    elif request.param == "small_with_newlines_blob":
+        return request.getfixturevalue("small_with_newlines_uploaded_blob")
```
kyleknap, line 87: We should be able to make this a bit more programmatic and simplify down to:
```python
@pytest.fixture
def uploaded_blob(request):
    return request.getfixturevalue(f"{request.param}_uploaded_blob")
```
The nice part is as long as we keep the `_uploaded_blob` naming pattern for our fixtures this fixture should automatically support any new blob fixtures we make. This also allows our parameterization to be less verbose by being able to specify `small`, `large`, etc. instead of the full name of the fixture.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
 
 
 def random_resource_name(name_length=8):
-    return "".join(random.choices(string.ascii_lowercase + string.digits, k=name_length))
+    return "".join(
+        random.choices(string.ascii_lowercase + string.digits, k=name_length)
+    )
+
+
+def sample_data(data_length=20, newline=False):
```
kyleknap, line 96: For the `sample_data` function, it may make sense to keep it as is and then we can delegate it for random data. Mainly for generating random newline data we will probably want more granularity such as how long we want each line and how many newlines we want. We can probably just create a new function called `sample_data_with_newlines()` and it can accept a `line_length` and `num_lines` and from there we can call `sample_data` by the number of lines and join with newlines so that we can guarantee that we are getting newlines in our content.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
+    @pytest.mark.parametrize(
+        "uploaded_blob",
+        [
+            "small_blob",
+            "large_blob",
+        ],
+        indirect=True,
+    )
+    def test_reads_all_data(self, uploaded_blob):
+        blob = uploaded_blob
```
kyleknap, line 114: Let's avoid renaming the fixture variable in the test unless it is necessary. Mainly it helps keep the test bodies small and easy to read. One option that we can explore if we don't like the `uploaded_blob` terminology is we can just rename the fixtures to `blob` and `<qualifier>_blob`, removing the `uploaded` part? I might like that more than my previous suggestion of including `uploaded` if you think that will help with following the logic. Thoughts?

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
+        with BlobIO(blob.url, "rb") as f:
+            assert f.read() == blob.data
+
+    @pytest.mark.parametrize("n", [1, 2, 4, 5, 9])
```
kyleknap, line 118: For the parametrization of both the read_n and seek cases, I think we can cut it down to the following cases:
1. Use `1` for the smallest possible amount that does not result in no-op
2. Use an amount between `1` and the size of the blob. We could choose 4, 5, or 9 here
3. A size that is equal to size of the blob
4. A size that is greater than the blob
The main reason that I'm suggesting it is that for each parameterization there is a logical case that we are trying to handle and it helps reduce cases that sort of overlap in what they are asserting and don't add much additional value.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
+    @pytest.mark.parametrize("uploaded_blob", ["small_blob"], indirect=True)
+    def test_random_seeks_and_reads(self, uploaded_blob, n):
+        blob = uploaded_blob
+        with BlobIO(blob.url, "rb") as f:
+            f.seek(n)
+            assert f.read() == blob.data[n:]
+
+    @pytest.mark.parametrize(
+        "uploaded_blob", ["small_with_newlines_blob"], indirect=True
+    )
```
kyleknap, line 136: For now, let's go ahead and remove `parameterization` of test cases that only include a single blob value and just use the specific fixture directly (e.g. `small_with_newlines_uploaded_blob`). Mainly it helps with readability and we can always add this back if there are more blobs that we want to add.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
+    def test_read_using_iter(self, uploaded_blob):
+        blob = uploaded_blob
+        with BlobIO(blob.url, "rb") as f:
+            data = b""
```
kyleknap, line 140: For the iter test, it would be interesting to just collect each individual line in a list and we append each iteration to this list and from there we can just compare against splitting of the original data on newlines. The reason I'm suggesting it is that it is a stronger assertion that we are only returning a line for iteration of the for loop and we are currently only looking at the entire data content.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,100 @@ def container_client(account_url):
...
+        [
+            "small_blob",
+            "large_blob",
+        ],
+        indirect=True,
+    )
+    def test_reads_all_data(self, uploaded_blob):
+        blob = uploaded_blob
+        with BlobIO(blob.url, "rb") as f:
+            assert f.read() == blob.data
```
kyleknap, line 116: For any reads, it is probably worth calling `tell()` and assert that we are in the expected position after a read. This is mainly an additional assurance that the position that we are telling a customer using the `tell()` interface is correct.

kyleknap (commented):
This is looking good! It's great to see that we cut the test time in half with these changes. I just had some more smaller suggestions, but we should be set after this.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,87 @@ def container_client(account_url):
...
+            assert f.read() == small_blob.data[n:]
+            if n < len(small_blob.data):
+                assert f.tell() == len(small_blob.data)
+
+    def test_read_using_iter(self, small_with_newlines_blob):
+        with BlobIO(small_with_newlines_blob.url, "rb") as f:
+            data = f.read().splitlines()
+            expected_data = small_with_newlines_blob.data.splitlines()
+            for i in data:
+                assert data == expected_data
```
kyleknap, line 130: For the iter test, let's update it to this:
```python
    def test_read_using_iter(self, small_with_newlines_blob):
        with BlobIO(small_with_newlines_blob.url, "rb") as f:
            lines = [line for line in f]
            expected_lines = small_with_newlines_blob.data.splitlines(keepends=True)
            assert lines == expected_lines
```
The main things are that:
1. We still want to be using the iter on the `BlobIO` object since that is custom to this library
2. This one is more subtle but it is a stronger assertion to assert the entire list instead of doing it one at a time because the `data` could be a subset of `expected_data` and still pass the test even though it is not the same content

tests/e2e/test_read.py
```diff
@@ -39,29 +44,87 @@ def container_client(account_url):
...
+        with BlobIO(blob.url, "rb") as f:
+            assert f.read() == blob.data
+            assert f.tell() == len(blob.data)
+
+    @pytest.mark.parametrize("n", [1, 5, 20, 21])
+    def test_read_n_bytes(self, small_blob, n):
+        with BlobIO(small_blob.url, "rb") as f:
+            for i in range(0, len(small_blob.data), n):
+                assert f.read(n) == small_blob.data[i : i + n]
+                if n < len(small_blob.data):
```
kyleknap, line 114: Let's avoid if statements related to assertion making; it can cause unintentional gaps in our testing if the if statement does not behave as expected. Generally, if we find ourselves needing if statements, it is a sign that we may need a separate test case. In this case, we should be able to update the assertion to something like this:
```python
                assert f.read(n) == small_blob.data[i : i + n]
                expected_position = min(i + n, len(small_blob.data))
                assert f.tell() == expected_position
```
to enforce the assertion that the tell won't go past the end of the blob.

tests/e2e/test_read.py
```diff
@@ -39,29 +44,87 @@ def container_client(account_url):
...
+                    assert f.tell() == i + n
+
+    @pytest.mark.parametrize("n", [1, 5, 20, 21])
+    def test_random_seeks_and_reads(self, small_blob, n):
+        with BlobIO(small_blob.url, "rb") as f:
+            f.seek(n)
+            assert f.read() == small_blob.data[n:]
+            if n < len(small_blob.data):
+                assert f.tell() == len(small_blob.data)
```
kyleknap, line 123: Same thing here. We should be able to update it to:
```python
            f.seek(n)
            assert f.read() == small_blob.data[n:]
            expected_position = max(n, len(small_blob.data))
            assert f.tell() == expected_position
```

kyleknap (commented):
Looks good! Just a few more comments and we should be good to merge.

kyleknap (approved):
Looks great! :ship:

PR #14, opened 2025-03-13
Implement write for `BlobIO`

kyleknap:
When in write mode, the `BlobIO` has the following behavior:
* It is non-seekable
* Only supports uploading as block blobs
* Calls to `write()` are processed asynchronously; they immediately return and any errors are propagated in subsequent calls to `BlobIO` methods. This is an optimization to speed up `torch.save()` to avoid waiting for both the serialization of a model property and the actual staging of data.
* Small writes are cached in memory up to the stage block partition currently at 32 MiB
* Calls to `flush()` always stages any data cached in memory and waits for the stage block completion. This allows for synchronous processing of writes if needed.
* If any stage block operations result in an error, the error is considered fatal and any data that has been provided to the instance of `BlobIO` will not be committed. Any retrying of writes must be done using a new `BlobIO` instance.

### DIFF
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -35,11 +35,7 @@ SDK_CREDENTIAL_TYPE = Optional[
     ]
 ]
 SUPPORTED_WRITE_BYTES_LIKE_TYPE = Union[bytes, bytearray, memoryview]
-try:
-    STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future[str]
-except TypeError:
-    # Python <3.9 does not support generic typing for futures
-    STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future
+STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future[str]
 
 
 class AzStorageTorchBlobClient:
@@ -122,10 +118,12 @@ class AzStorageTorchBlobClient:
         # the max worker calculation from the executor class and inject it into both the executor
         # and semaphore
         #
-        # Note: Once we support Python 3.13, the default calculation for max workers for the executor
-        # uses os.process_cpu_count() instead. We should make sure to update the value here to match
-        # that as well for Python 3.13 environments.
-        return min(32, (os.cpu_count() or 1) + 4)
+        # In Python 3.13, os.process_cpu_count() was added and the ThreadPoolExecutor updated to
+        # use os.process_cpu_count() instead of os.cpu_count() when calculating default max workers.
+        # To match ThreadPoolExecutor defaults across Python versions, we use process_cpu_count
+        # if available, otherwise fall back to os.cpu_count().
+        cpu_count_fn = getattr(os, "process_cpu_count", os.cpu_count)
+        return min(32, (cpu_count_fn() or 1) + 4)
 
     @functools.cached_property
     def _blob_properties(self) -> azure.storage.blob.BlobProperties:
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -9,9 +9,12 @@ import functools
 import io
 import logging
 import math
+import os
 import random
+import threading
 import time
-from typing import Optional, List, Tuple, Iterator, Union
+import uuid
+from typing import Optional, List, Tuple, Iterator, Union, Type
 
 from azure.core.credentials import (
     AzureSasCredential,
@@ -31,6 +34,12 @@ SDK_CREDENTIAL_TYPE = Optional[
         TokenCredential,
     ]
 ]
+SUPPORTED_WRITE_BYTES_LIKE_TYPE = Union[bytes, bytearray, memoryview]
+try:
+    STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future[str]
+except TypeError:
+    # Python <3.9 does not support generic typing for futures
+    STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future
 
 
 class AzStorageTorchBlobClient:
@@ -38,6 +47,7 @@ class AzStorageTorchBlobClient:
     _PARTITIONED_DOWNLOAD_THRESHOLD = 16 * 1024 * 1024
     _PARTITION_SIZE = 16 * 1024 * 1024
     _NUM_DOWNLOAD_ATTEMPTS = 3
+    _STAGE_BLOCK_SIZE = 32 * 1024 * 1024
     _RETRYABLE_READ_EXCEPTIONS = (
         azure.core.exceptions.IncompleteReadError,
         azure.core.exceptions.HttpResponseError,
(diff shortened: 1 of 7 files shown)


src/azstoragetorch/io.py
```diff
@@ -229,9 +284,74 @@ def _compute_new_position(self, offset: int, whence: int) -> int:
...
+        block_ids = [f.result() for f in self._all_stage_block_futures]
+        self._raise_if_duplicate_block_ids(block_ids)
+        self._client.commit_block_list(block_ids)
+
+    def _raise_if_duplicate_block_ids(self, block_ids: str) -> None:
+        # An additional safety measure to ensure we never reuse block IDs within a BlobIO instance. This
+        # should not be an issue with UUID4 for block IDs, but that may not always be the case if
+        # block ID generation changes in the future.
+        if len(block_ids) != len(set(block_ids)):
+            raise RuntimeError(
```
jalauzon-msft, line 319: Do we want to raise `FatalBlobIOWriteError` instead?

src/azstoragetorch/io.py
```diff
@@ -229,9 +284,74 @@ def _compute_new_position(self, offset: int, whence: int) -> int:
...
+        self._process_stage_block_futures_for_errors(wait)
+
+    def _process_stage_block_futures_for_errors(self, wait: bool) -> None:
+        if wait:
+            concurrent.futures.wait(
+                self._in_progress_stage_block_futures,
+                return_when=concurrent.futures.FIRST_EXCEPTION,
+            )
+        futures_still_in_progress = []
+        for future in self._in_progress_stage_block_futures:
```
jalauzon-msft, line 337: So essentially this is saving the FIRST exception? Do we care about the others? I guess maybe not.

src/azstoragetorch/io.py
```diff
@@ -229,9 +284,74 @@ def _compute_new_position(self, offset: int, whence: int) -> int:
...
+            if future.done():
+                if (
+                    self._stage_block_exception is None
+                    and future.exception() is not None
+                ):
+                    self._stage_block_exception = future.exception()
+            else:
+                futures_still_in_progress.append(future)
+        self._in_progress_stage_block_futures = futures_still_in_progress
```
jalauzon-msft, line 346: Is there a mechanism to, and would we want to, cancel all the pending futures if we detect an error since we know ultimately this thing won't commit?

src/azstoragetorch/_client.py
```diff
@@ -31,13 +34,20 @@
         TokenCredential,
     ]
 ]
+SUPPORTED_WRITE_BYTES_LIKE_TYPE = Union[bytes, bytearray, memoryview]
+try:
+    STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future[str]
+except TypeError:
+    # Python <3.9 does not support generic typing for futures
```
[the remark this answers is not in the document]
kyleknap: This is a good callout. I forgot Python 3.8 was going EoL right when we originally setup our Python version matrix for the project. So, I did not think much about it here.

Looking at the latest version of PyTorch: https://pypi.org/project/torch/2.6.0/, they already dropped support for 3.8 and added support for 3.13. I propose we do the same as the intention was to follow PyTorch's support matrix. I can send this up as a separate pull request. Once merged, I can rebase this and remove the try/catch.

src/azstoragetorch/_client.py
```diff
@@ -76,9 +95,38 @@ def download(self, offset: int = 0, length: Optional[int] = None) -> bytes:
...
     def close(self) -> None:
         self._executor.shutdown()
 
+    def _get_max_in_flight_requests(self) -> int:
+        # Ideally we would just match this value to the max workers of the executor. However
+        # the executor class does not publicly expose its max worker count. So, instead we copy
+        # the max worker calculation from the executor class and inject it into both the executor
+        # and semaphore
+        #
+        # Note: Once we support Python 3.13, the default calculation for max workers for the executor
```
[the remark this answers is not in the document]
kyleknap: Building on this [suggestion](https://github.com/Azure/azure-storage-for-pytorch/pull/14#discussion_r2006598698), I'll go ahead and add the compatibility logic. I didn't do it this time because it felt awkward to add logic that was not tested because we did not support 3.13 in our test matrix.

src/azstoragetorch/_client.py
```diff
@@ -76,9 +95,38 @@ def download(self, offset: int = 0, length: Optional[int] = None) -> bytes:
...
+        self, data: SUPPORTED_WRITE_BYTES_LIKE_TYPE
+    ) -> List[STAGE_BLOCK_FUTURE_TYPE]:
+        if not data:
+            raise ValueError("Data must not be empty.")
+        stage_block_partitions = self._get_stage_block_partitions(data)
+        futures = []
+        for pos, length in stage_block_partitions:
+            self._max_in_flight_semaphore.acquire()
+            future = self._executor.submit(self._stage_block, data[pos : pos + length])
+            future.add_done_callback(self._release_in_flight_semaphore)
```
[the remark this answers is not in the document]
kyleknap: Yep any done callbacks will run no matter the result (success, error, or cancelled). They are even run if the future is completed before the `add_done_callback()` is called, in which case it will just be ran immediately in the thread calling `add_done_callback()`

src/azstoragetorch/io.py
kyleknap: I was on the fence about this one when I chose `RuntimeError`. I could go either way especially since if a customer ever got this error, the best option (assuming there is not consistent collisions) is to just retry it. I went with `RuntimeError` to be more indicative that this was not expected and should never be hit (i.e. panic as this is definitely a bug). Thoughts? I'm good with changing it if you like that more.

src/azstoragetorch/io.py
kyleknap: Yep only the first one is saved. I mainly chose to do that because I'm not sure how valuable is to have all of the exceptions and the error message it could potentially get overwhelming if all simultaneous stage blocks hit the same error (e.g. permissions error) and we displayed them all. Thoughts?

src/azstoragetorch/io.py
kyleknap: So future's have a [`cancel()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel) mechanism. Unfortunately, it is really only helpful if we have a bunch of futures queued up (e.g. sitting in a list somewhere in the executor but not actively being worked on by a thread) because cancelling will just prevent the queued ones from executing, but the ones actively being uploaded cannot be interrupted. We would have to introduce our own mechanism to interrupt a thread mid upload, which would be tricky given we are passing bytes-like objects.

So, because we bound the number of in-progress futures to the number of worker threads, we really will not have any futures in this "queued" state and therefore would not be able to `cancel()` them and have them stop immediately.

src/azstoragetorch/_client.py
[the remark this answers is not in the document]
jalauzon-msft: Sounds good then, thanks.

src/azstoragetorch/io.py
jalauzon-msft: I didn't have a strong opinion so this is fine.

src/azstoragetorch/io.py
jalauzon-msft: Ok, yeah, I think that makes sense and agree showing all exceptions may be a bit too much information as they are likely all the same. I also don't know a great way to show them if we don't have something like AggregateException in .NET. This is probably fine for now and if we get some feedback on it, we can adjust.

src/azstoragetorch/io.py
jalauzon-msft: Ok, makes sense.

src/azstoragetorch/io.py
kyleknap: Sounds good. Yeah I agree if we get feedback that customers want all of the exceptions, we should be able to add it. Just need to make sure the end message is still readable.

jalauzon-msft (approved):

PR #15, opened 2025-03-14
Test for loading a pytorch model

anjaliratnam-msft:
Checks that a pytorch model can be loaded using BlobIO() and torch.load().


tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+
+class TestLoad:
+    def test_load_existing_model(self, blob_url, model):
+        test_model = torch.hub.load(
+            "pytorch/vision:v0.10.0", "resnet101", pretrained=False
+        )
+
+        with BlobIO(blob_url, "rb") as f:
+            state_dict = torch.load(f)
+            test_model.load_state_dict(state_dict)
```
kyleknap, line 74: We should see if we need to explicitly load the state dict into the model or if we can just compare the `state_dict` taken directly from `torch.load()` with our original model state_dict. I'm mainly suggesting this because it will allows us to avoid doing another call out to `torch.hub`.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for key in model.state_dict():
```
kyleknap, line 76: Before iterating through the keys, we should make sure the keys in both state dictionaries are the same. Otherwise, if `model` is a subset of `test_model`, this test will unexpectedly pass.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+            assert torch.equal(test_model.state_dict()[key], model.state_dict()[key])
```
kyleknap, line 77: When iterating through the state_dict, it may be worth calling `items()` on the state_dict to get the direct value and help make this line a bit less verbose.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+    blob_name = "model.pth"
+    torch.save(model.state_dict(), blob_name)
+    blob_client = container_client.get_blob_client(blob=blob_name)
+    with open(blob_name, "rb") as f:
+        blob_client.upload_blob(f)
+    return blob_name
+
+
+@pytest.fixture(scope="module")
+def account_url():
```
kyleknap, line 35: Instead of writing the fixture again, let's move it out to a [`conftest.py`](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files) file so that we can share it between `test_load.py` and `test_read.py`.

Also, we will probably want to expand these shared fixtures to the `package` scope, instead of `module`, so that we use the same `account_url` and `container_client` for all end to end tests when we do this.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+from azstoragetorch.io import BlobIO
+from azure.identity import DefaultAzureCredential
+from azure.storage.blob import BlobServiceClient
+
+
+@pytest.fixture(scope="module")
+def model():
+    ssl._create_default_https_context = functools.partial(
+        ssl.create_default_context, cafile=certifi.where()
+    )
```
kyleknap, line 19: For this line, we should probably do the following:

1. Move it out to it's own fixture. We can give it the name `ssl_inject_certifi`
2. In the fixture, use [`monkeypatch`](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) built-in fixture to make sure we restore state back to the original value. Mainly in doing this we are mutating global state and when global state is mutated it can impact logic later on and hard to tell why it is being impacted in a particular way. So we want to be careful whenever we do things like this.
3. Add a comment in the fixture on why we need this. Mainly `torch.hub` does not use `certifi` which means it delegates to system certificates and some environments may run into SSL errors without it.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,77 @@
...
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
```
kyleknap, line 20: When we call out to `torch.load`, we should probably make sure to download it to the non-default hub directory for the following reasons:
* It possibly makes the test possibly not consistently reproducible if one environment has a cached version that differs from another 
* Tests generally should completely clean up after themselves and in this case we are leaving around cached models possibly
* In future CI environments, we may not have access to write in the home directory.

To address it, we should probably set one of the options [here](https://pytorch.org/docs/stable/hub.html#where-are-my-downloaded-models-saved) and set it to a temporary directory using the [`tmp_path`](https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-tmp_path) fixture.

kyleknap (commented):
Looks good! I just had some general refactoring suggestions but I like the approach we went with.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,60 @@
...
+def model(tmp_path):
+    torch.hub.set_dir(tmp_path)
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
+    return model
+
+
+@pytest.fixture(scope="module")
+def upload_model(model, container_client):
+    blob_name = "model.pth"
+    torch.save(model.state_dict(), blob_name)
```
kyleknap, line 40: We should probably use the [`tmp_path`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.tmpdir.tmp_path) fixture for the name of the file we save the model to. Right now it looks like the `model.pth` is being saved to the current working directory of our tests, which means it won't be deleted after cleanup.

Another option is we can just provide a [`BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO) object to `torch.save()` and use that directly when uploading. The nice part about this is that we would not be writing to disk and possibly leak state between test runs. I don't have a strong preference though on which option we go with.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,60 @@
...
+import torch
+import ssl
+import functools
+import certifi
+
+from azstoragetorch.io import BlobIO
+
+
+@pytest.fixture(scope="module")
+def ssl_inject_certifi(monkeypatch):
```
kyleknap, line 16: Are the test currently passing for your environment? I'm not seeing the fixture being referenced anywhere so I'd assume it would not be applied meaning we would be hitting the issues we were seeing in your environment. If they are passing, we should just remove the fixture as that means both of us are fine running the tests without it. We can add it back if the problem comes up again.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,60 @@
...
+    blob_name = "model.pth"
+    torch.save(model.state_dict(), blob_name)
+    blob_client = container_client.get_blob_client(blob=blob_name)
+    with open(blob_name, "rb") as f:
+        blob_client.upload_blob(f)
+    return blob_name
+
+
+@pytest.fixture()
+def blob_url(account_url, container_client, upload_model):
```
kyleknap, line 48: Could we inject the word `state_dict` into this fixture name (e.g. `state_dict_blob_url`)? Mainly it makes it a little more clear that the blob url refers to the url to the model's state dict.

kyleknap (commented):
Looks good! I just had some follow up comments and we should be set here.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.fixture(scope="module", autouse=True)
+def torch_hub_cache(tmp_path_factory):
+    current_dir = torch.hub.get_dir()
+    yield tmp_path_factory.mktemp("torch_hub")
+    torch.hub.set_dir(current_dir)
+
+
+@pytest.fixture(scope="module")
+def model(torch_hub_cache):
+    torch.hub.set_dir(torch_hub_cache)
```
kyleknap, line 21: Ah so for this `set_dir` call, we should move it to the `torch_hub_cache` fixture so that we set it as part of the fixture prior to yielding.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,49 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def torch_hub_cache(tmp_path_factory):
+    current_dir = torch.hub.get_dir()
+    yield tmp_path_factory.mktemp("torch_hub")
+    torch.hub.set_dir(current_dir)
+
+
+@pytest.fixture(scope="module")
+def model(torch_hub_cache):
```
kyleknap, line 20: Now that we are doing `autouse` we should be able to remove reference to the `torch_hub_cache` here.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,49 @@
...
+    torch.hub.set_dir(torch_hub_cache)
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
+    return model
+
+
+@pytest.fixture(scope="module")
+def upload_model(model, container_client, torch_hub_cache):
+    blob_name = torch_hub_cache / "model.pth"
```
kyleknap, line 28: Instead of using `torch_hub_cache`, let's use `tmp_path` directly instead. My main concern in using the cache directly to save models is that `torch.hub` may be expecting a specific type of directory structure for its local cache and we may cause issues at some point in directly writing to it. By writing to a file outside of the torch hub cache, we eliminate that possibility.

kyleknap (commented):
Looks good. Just had a few more small comments.

anjaliratnam-msft:
Created new pr using a new branch

PR #18, opened 2025-03-31
Loading pytorch model (updated branch)

anjaliratnam-msft:
Checks that a pytorch model can be loaded using BlobIO() and torch.load(). Fixes test_read_using_iter().

### DIFF
--- a/tests/e2e/test_load.py
+++ b/tests/e2e/test_load.py
@@ -1,0 +1,51 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import pytest
+import torch
+from azstoragetorch.io import BlobIO
+
+
+@pytest.fixture(scope="module", autouse=True)
+def torch_hub_cache(tmp_path_factory):
+    current_dir = torch.hub.get_dir()
+    torch.hub.set_dir(tmp_path_factory.mktemp("torch_hub"))
+    yield
+    torch.hub.set_dir(current_dir)
+
+
+@pytest.fixture(scope="module")
+def model():
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
+    return model
+
+
+@pytest.fixture(scope="module")
+def tmp_path(tmp_path_factory):
+    return tmp_path_factory.mktemp("model")
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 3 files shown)


tests/e2e/test_load.py
```diff
@@ -0,0 +1,51 @@
...
+
+@pytest.fixture(scope="module")
+def model():
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
+    return model
+
+
+@pytest.fixture(scope="module")
+def tmp_path(tmp_path_factory):
+    return tmp_path_factory.mktemp("model")
```
kyleknap, line 27: We should be able to remove this fixture completely and just use the built-in `tmp_path`. According to `py.test` the built-in [`tmp_path`](https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-tmp_path) will create a unique temporary directory for each invocation.

tests/e2e/test_load.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.fixture(scope="module")
+def upload_model(model, container_client, tmp_path):
+    blob_name = tmp_path / "model.pth"
```
kyleknap, line 32: Let's rename `blob_name` to `model_path`. Mainly it is a little confusing that this variable really only points to a path to a local model and is not actually the name of the blob. So that would help make the variable a bit more representative of what it's value is.

tests/e2e/test_read.py
```diff
@@ -125,5 +99,8 @@ def test_random_seeks_and_reads(self, small_blob, n):
     def test_read_using_iter(self, small_with_newlines_blob):
         with BlobIO(small_with_newlines_blob.url, "rb") as f:
             lines = [line for line in f]
-            expected_lines = small_with_newlines_blob.data.splitlines(keepends=True)
+            expected_lines = [
+                line + b"\n" for line in small_with_newlines_blob.data.split(b"\n")[:-1]
+            ]
+            expected_lines.append(small_with_newlines_blob.data.split(b"\n")[-1])
```
kyleknap, line 105: For fixing this test, let's just wrap the expected data with a `io.BytesIO` object to get the expected lines e.g.:
```python
 expected_lines = io.BytesIO(small_with_newlines_blob.data).readlines()
```
Mainly this helps make it less verbose and easier to follow what we are expecting.

kyleknap (commented):
Looks good. I just had a few comments and we should be set here.

kyleknap (approved):
Looks good! :ship:

PR #19, opened 2025-04-01
Test for saving a model

anjaliratnam-msft:
Tests for saving a model using BlobIO and torch.save.


tests/e2e/test_save.py
```diff
@@ -0,0 +1,43 @@
...
+    current_dir = torch.hub.get_dir()
+    torch.hub.set_dir(tmp_path_factory.mktemp("torch_hub"))
+    yield
+    torch.hub.set_dir(current_dir)
+
+
+@pytest.fixture(scope="module")
+def model():
+    model = torch.hub.load("pytorch/vision:v0.10.0", "resnet101", pretrained=False)
+    return model
```
kyleknap, line 23: It probably makes sense to move both of these fixtures to the `conftest.py` file so that we can reuse it between the `test_save.py` and `test_load.py` files. We'll probably have to expand the scope to package for these as well.

Another option that I like as well is if we combine the `test_save.py` and `test_load.py` into a `test_torch_serialize.py` file where we generally test integration with pytorch serialization methods. I'm fine with either option.

tests/e2e/test_save.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@pytest.fixture(scope="module")
+def state_dict_blob_url(account_url, container_client, tmp_path_factory):
+    model_name = tmp_path_factory.mktemp("model") / "model.pth"
```
kyleknap, line 28: It might make sense to make the blob name randomized here since we already use this name for the torch load versions of the tests. Mainly by randomizing the name here, we guarantee that we are writing to the blob from a clean slate and we do not run into situations where even if the write failed, the tests pass because a previous fixture had uploaded the model already.

tests/e2e/test_save.py
```diff
@@ -0,0 +1,43 @@
...
+        with BlobIO(state_dict_blob_url, "wb") as f:
+            torch.save(model.state_dict(), f)
+
+        with BlobIO(state_dict_blob_url, "rb") as f:
+            state_dict = torch.load(f)
+
+        assert state_dict.keys() == model.state_dict().keys()
+
+        for key, value in model.state_dict().items():
+            assert torch.equal(state_dict[key], value)
```
kyleknap, line 43: It might make sense to turn this into a utility function `assert_state_dict()` that compares two state dicts and we can reuse for both the load and the save methods.

kyleknap (commented):
Looks good. I just had a few refactoring suggestions.

tests/e2e/test_torch_serialize.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.fixture(scope="module", autouse=True)
+def upload_model(model, container_client, model_path_name):
+    torch.save(model.state_dict(), model_path_name)
+    blob_client = container_client.get_blob_client(blob=model_path_name.name)
+    with open(model_path_name, "rb") as f:
+        blob_client.upload_blob(f)
+
+
+@pytest.fixture()
+def state_dict_blob_url(account_url, container_client, model_path_name):
```
kyleknap, line 42: I still think we may be leaving state between test cases because even though the `model_path_name` uses a random resource name, the fixture is set to `module` scope so it will be only chosen randomly once. Instead I'm thinking we:
1. Introduce a `container_url` fixture that returns the container url
2. Use the `container_url` fixture for the `state_dict_blob_url` fixture
3. For the `test_save_model` method, we use the `container_url` fixture instead of `state_dict_blob_url` and generate a new random resource name in the test body to guarantee we will be writing to a blob that did not previously exist nor be reused in other tests.

kyleknap (commented):
Looks good. I just had one more comment and we should be good to get this merged

kyleknap (approved):
Looks good! :ship:

PR #20, opened 2025-04-01
Tests for write function

anjaliratnam-msft:
Tests for a single write, multiple writes, and raising the FatalBlobIOError.

### DIFF
--- a/tests/e2e/test_write.py
+++ b/tests/e2e/test_write.py
@@ -41,7 +41,6 @@ def container_client(account_url):
     )
     container_name = random_resource_name()
     container = blob_service_client.create_container(name=container_name)
-    container.set_container_access_policy(signed_identifiers={}, public_access=None)
     yield container
     container.delete_container()
 
@@ -121,8 +120,8 @@ class TestWrite:
         indirect=True,
     )
     def test_write_error(self, blob):
-        with BlobIO(blob.url, "wb", credential=None) as f:
-            with pytest.raises(FatalBlobIOWriteError):
+        with pytest.raises(FatalBlobIOWriteError):    
+            with BlobIO(blob.url, "wb", credential=False) as f:
                 f.write(blob.data)
                 f.flush()
             assert not f.closed
--- a/tests/e2e/test_write.py
+++ b/tests/e2e/test_write.py
@@ -3,16 +3,13 @@
 # Licensed under the MIT License. See LICENSE in the project root for
 # license information.
 # --------------------------------------------------------------------------
-from dataclasses import dataclass
 import os
 import random
 import string
-
 import pytest
-from azstoragetorch.io import BlobIO
 
-from azure.identity import DefaultAzureCredential
-from azure.storage.blob import BlobServiceClient
+from dataclasses import dataclass
+from azstoragetorch.io import BlobIO
 from azstoragetorch.exceptions import FatalBlobIOWriteError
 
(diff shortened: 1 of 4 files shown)


tests/e2e/test_write.py
```diff
@@ -0,0 +1,103 @@
...
+
+def write_blob_url(account_url, container_client, data):
+    blob_name = random_resource_name()
+    url = f"{account_url}/{container_client.container_name}/{blob_name}"
+
+    return Blob(data=data, url=url)
+
+
+def sample_data(data_length=20):
+    return os.urandom(data_length)
```
kyleknap, line 55: It might make sense to try to consolidate the various versions that we have of `sample_data` and `random_resource_name` functions so that we can reuse it. I'm thinking we can introduce a `utils.py` file for end to end tests similar to what I did here for the unit tests: https://github.com/Azure/azure-storage-for-pytorch/blob/main/tests/unit/utils.py and we can then import directly from the `utils.py` for any shared end to end test utilities.

tests/e2e/test_write.py
```diff
@@ -0,0 +1,103 @@
...
+            ("large", _STAGE_BLOCK_SIZE * 2),
+            ("large", _STAGE_BLOCK_SIZE * 3),
+        ],
+        indirect=["blob"],
+    )
+    def test_write_content_split_up(self, blob, n):
+        with BlobIO(blob.url, "wb") as f:
+            for i in range(0, len(blob.data), n):
+                assert f.write(blob.data[i : i + n]) == min(n, len(blob.data))
+            assert f.tell() == len(blob.data)
```
kyleknap, line 88: It would be good to assert the `tell()` position for each write iteration as well to make sure `tell()` is current as we write each chunk.

tests/e2e/test_write.py
```diff
@@ -0,0 +1,103 @@
...
+            ("small", 21),
+            ("large", _STAGE_BLOCK_SIZE * 2),
+            ("large", _STAGE_BLOCK_SIZE * 3),
+        ],
+        indirect=["blob"],
+    )
+    def test_write_content_split_up(self, blob, n):
+        with BlobIO(blob.url, "wb") as f:
+            for i in range(0, len(blob.data), n):
+                assert f.write(blob.data[i : i + n]) == min(n, len(blob.data))
```
kyleknap, line 87: It probably makes sense to just assign the chunk to a variable e.g. `chunk = blob.data[i : i +n]` so that we can just assert against the length of the chunk directly for the expected return value from `write()`.

tests/e2e/test_write.py
```diff
@@ -0,0 +1,103 @@
...
+            ("small", 1),
+            ("small", 5),
+            ("small", 20),
+            ("small", 21),
+            ("large", _STAGE_BLOCK_SIZE * 2),
+            ("large", _STAGE_BLOCK_SIZE * 3),
+        ],
+        indirect=["blob"],
+    )
+    def test_write_content_split_up(self, blob, n):
```
kyleknap, line 84: Instead of `split_up` let's change the terminology to `chunks` so `test_write_content_in_chunks` for the test name. Mainly that is a little more clear what the intention of the test is.

kyleknap (commented):
Looks good. Just had some comments related to refactoring as we expand the end to end test suite.

kyleknap (approved):
Looks great! :ship:

PR #21, opened 2025-04-03
Updated user agent

anjaliratnam-msft:
Added azstoragetorch/0.0.1 to the user agent and added a unit test for it.

### DIFF
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -24,6 +24,7 @@ import azure.core.exceptions
 import azure.storage.blob
 import azure.storage.blob._generated.models
 from azure.storage.blob._shared.response_handlers import process_storage_error
+from azstoragetorch._version import __version__
 
 
 _LOGGER = logging.getLogger(__name__)
@@ -58,8 +59,7 @@ class AzStorageTorchBlobClient:
     ):
         self._sdk_blob_client = sdk_blob_client
         self._generated_sdk_storage_client = self._sdk_blob_client._client
-        self._sdk_blob_client._config.user_agent_policy.add_user_agent("azstoragetorch/0.0.1")
-
+        
         if max_in_flight_requests is None:
             max_in_flight_requests = self._get_max_in_flight_requests()
         if executor is None:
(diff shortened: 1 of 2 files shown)


src/azstoragetorch/_client.py
```diff
@@ -58,6 +58,8 @@ def __init__(
     ):
         self._sdk_blob_client = sdk_blob_client
         self._generated_sdk_storage_client = self._sdk_blob_client._client
+        self._sdk_blob_client._config.user_agent_policy.add_user_agent("azstoragetorch/0.0.1")
```
kyleknap, line 61: For the version value, we should try to import if from [`_version.py`](https://github.com/Azure/azure-storage-for-pytorch/blob/490658e58084dd06e8c078d9ac3cadd45810d7f9/src/azstoragetorch/_version.py#L7) so the user agent is automatically updated with bump of the library's version.

kyleknap, line 61: In general, we should avoid accessing properties that begin with underscore (e.g. `_config`) because they are not considered public interfaces and our logic could break unexpectedly on an upgrade of the Python SDK.

Instead, we should try to provide the user agent as a keyword argument when we construct the client (e.g. `from_blob_url` method on the SDK client) as that will be a public interface. There is not great documentation on how to do it, but it is hinted at here: https://azure.github.io/azure-sdk/python_implementation.html#service-client. I think the keyword argument that we will need is `user_agent` per this class: https://github.com/Azure/azure-sdk-for-python/blob/c83193dd1391b27b283d0e7c07816a84d51cca1c/sdk/core/azure-core/azure/core/pipeline/policies/_universal.py#L209. That looks like it maps to `application_id`.

kyleknap (commented):
Looks good. I just had a couple of small comments.

src/azstoragetorch/_client.py
jalauzon-msft: Yes, this would be how I recommend doing this, using the `user_agent` keyword on client construction. This should automatically be pre-pended to the SDK user agent.

tests/unit/test_client.py
```diff
@@ -59,6 +60,7 @@ def mock_generated_sdk_storage_client():
 def mock_sdk_blob_client(mock_generated_sdk_storage_client):
     mock_sdk_client = mock.Mock(BlobClient)
     mock_sdk_client._client = mock_generated_sdk_storage_client
+    mock_sdk_client._config = mock.Mock()
```
kyleknap, line 63: Let's remove this added mock now that we went with a different approach

kyleknap (commented):
Looks good. I just a small comment and we should be good to merge.

kyleknap (approved):
Looks good! :ship:

PR #22, opened 2025-04-07
Sphinx

anjaliratnam-msft:
Added sphinx to the project.


pyproject.toml
```diff
@@ -39,6 +39,8 @@ dev = [
     "build",
     "check-manifest",
     "pytest",
+    "sphinx",
```
kyleknap, line 42: We should also make sure to regenerate the `requirements-dev.txt` file. This is our lock file to make sure all of the dependencies that we use are the same across environments.

To regenerate it, you'll need a Python 3.9 installation and [`pip-tools`](https://github.com/jazzband/pip-tools?tab=readme-ov-file#installation) installed. Once installed, you should be able to run this [command](https://github.com/Azure/azure-storage-for-pytorch/blob/b352b084bdbd44f54f22432e5d0508d9f9e52e6c/requirements-dev.txt#L5) from the root of the directory to get the requirements file automatically updated.

kyleknap (commented):
This is looking good. Just posting my comments here as I had most of them done before the PR was closed. We can apply them to the new PR.

PR #28, opened 2025-04-17
Read optimization

anjaliratnam-msft:
Remove the call to get_blob_properties on the first read and get properties from the response header.

### DIFF
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -195,6 +195,7 @@ class AzStorageTorchBlobClient:
             max_in_flight_requests = self._get_max_in_flight_requests()
         self._max_in_flight_requests = max_in_flight_requests
         self._executor = executor
+        self._blob_properties = None
 
     @property
     def url(self) -> str:
@@ -213,14 +214,26 @@ class AzStorageTorchBlobClient:
         return self._sdk_blob_client.container_name
 
     def get_blob_size(self) -> int:
-        return self._blob_properties.size
+        return self._get_blob_properties().size
 
     def download(self, offset: int = 0, length: Optional[int] = None) -> bytes:
+        initial_content = b""
+        if self._blob_properties is None:
+            initial_content = self._download_from_unknown_blob_size(offset, length)
+            offset = len(initial_content) + offset
+            if length is not None:
+                length = length - len(initial_content)
+            if not self._more_to_download(offset, length):
+                return initial_content
         length = self._update_download_length_from_blob_size(offset, length)
         if length < self._PARTITIONED_DOWNLOAD_THRESHOLD:
-            return self._download_with_retries(offset, length)
+            return b"".join(
+                [initial_content, self._download_with_retries(offset, length)]
+            )
         else:
-            return self._partitioned_download(offset, length)
+            return b"".join(
+                [initial_content, self._partitioned_download(offset, length)]
+            )
 
     def stage_blocks(
         self, data: SUPPORTED_WRITE_BYTES_LIKE_TYPE
@@ -285,9 +298,10 @@ class AzStorageTorchBlobClient:
         # submitted to the executor when there are no workers available to upload it.
         return threading.Semaphore(self._max_in_flight_requests)
 
-    @functools.cached_property
-    def _blob_properties(self) -> azure.storage.blob.BlobProperties:
-        return self._sdk_blob_client.get_blob_properties()
+    def _get_blob_properties(self) -> azure.storage.blob.BlobProperties:
+        if self._blob_properties is None:
+            self._blob_properties = self._sdk_blob_client.get_blob_properties()
+        return self._blob_properties
 
     def _update_download_length_from_blob_size(
         self, offset: int, length: Optional[int] = None
@@ -323,6 +337,22 @@ class AzStorageTorchBlobClient:
             partitions.append((start, size))
         return partitions
 
+    def _more_to_download(
+        self, updated_offset, remaining_length: Optional[int] = None
+    ) -> bool:
+        if self._blob_properties.size <= updated_offset:
+            return False
+        if remaining_length is not None and remaining_length == 0:
+            return False
+        return True
+
+    def _download_from_unknown_blob_size(
+        self, offset: int, length: Optional[int] = None
+    ) -> bytes:
+        if length is None or length > self._PARTITIONED_DOWNLOAD_THRESHOLD:
+            length = self._PARTITIONED_DOWNLOAD_THRESHOLD
+        return self._download_with_retries(offset, length)
+
     def _download_with_retries(self, pos: int, length: int) -> bytes:
         attempt = 0
         while self._attempts_remaining(attempt):
@@ -342,20 +372,54 @@ class AzStorageTorchBlobClient:
                 )
                 time.sleep(backoff_time)
 
+    def _set_blob_properties_from_download(self, response) -> None:
+        headers = response.response.headers
+        blob_size = self._get_size_from_range(headers["Content-Range"])
+        self._blob_properties = azure.storage.blob.BlobProperties(
+            **{"Content-Length": blob_size, "ETag": headers.get("ETag")}
+        )
+
     def _get_download_stream(self, pos: int, length: int) -> Iterator[bytes]:
         try:
-            return self._generated_sdk_storage_client.blob.download(
-                range=f"bytes={pos}-{pos + length - 1}",
-                modified_access_conditions=azure.storage.blob._generated.models.ModifiedAccessConditions(
-                    if_match=self._blob_properties.etag
-                ),
+            download_kwargs = {
+                "range": f"bytes={pos}-{pos + length - 1}",
+            }
+            if self._blob_properties is not None:
+                download_kwargs["modified_access_conditions"] = (
+                    azure.storage.blob._generated.models.ModifiedAccessConditions(
+                        if_match=self._blob_properties.etag
+                    )
+                )
+            response = self._generated_sdk_storage_client.blob.download(
+                **download_kwargs
             )
+            if self._blob_properties is None:
+                self._set_blob_properties_from_download(response)
+            return response
         except azure.core.exceptions.HttpResponseError as e:
+            if self._is_invalid_range_from_empty_blob_error(e):
+                self._blob_properties = azure.storage.blob.BlobProperties(
+                    **{"Content-Length": 0}
+                )
+                return b""
             # TODO: This is so that we properly map exceptions from the generated client to the correct
             # exception class and error code. In the future, prior to a GA, we should consider pulling
             # in this function or a derivative of it if we plan to continue to raise Azure Python SDK
             # exceptions from this library (i.e. instead of raising our own exception classes).
             process_storage_error(e)
+
+    def _get_size_from_range(self, range_header: str) -> int:
+        return int(range_header.split(" ", 1)[1].split("/", 1)[1])
+
+    def _is_invalid_range_from_empty_blob_error(
+        self, error: azure.core.exceptions.HttpResponseError
+    ) -> bool:
+        return (
+            error.response
+            and error.status_code == 416
+            and "Content-Range" in error.response.headers
+            and self._get_size_from_range(error.response.headers["Content-Range"]) == 0
+        )
 
     def _attempts_remaining(self, attempt_number: int) -> int:
         return max(self._NUM_DOWNLOAD_ATTEMPTS - attempt_number, 0)
--- a/src/azstoragetorch/io.py
+++ b/src/azstoragetorch/io.py
@@ -86,6 +86,7 @@ class BlobIO(io.IOBase):
             _client.STAGE_BLOCK_FUTURE_TYPE
         ] = []
         self._stage_block_exception: Optional[BaseException] = None
+        self._blob_size = None
 
     def close(self) -> None:
         """Close the file-like object.
@@ -307,11 +308,15 @@ class BlobIO(io.IOBase):
         client_factory = _client.AzStorageTorchBlobClientFactory(credential=credential)
         return client_factory.get_blob_client_from_url(blob_url)
 
+    def _get_blob_size(self):
+        if self._blob_size is None:
+            self._blob_size = self._client.get_blob_size()
+        return self._blob_size
+
     def _readline(self, size: Optional[int]) -> bytes:
         consumed = b""
-        if size == 0 or self._is_at_end_of_blob():
+        if size == 0 or self._is_at_end_of_blob(fetch_blob_size=False):
             return consumed
-
         limit = self._get_limit(size)
         if self._readline_buffer:
             consumed = self._consume_from_readline_buffer(consumed, limit)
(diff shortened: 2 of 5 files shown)


src/azstoragetorch/io.py
```diff
@@ -258,7 +265,28 @@ def _should_download_more_for_readline(self, consumed: bytes, limit: int) -> boo
...
+    def _set_blob_size_from_header(self) -> int:
+        response = self._client._generated_sdk_storage_client.blob.download(
+            range=f"bytes={0}-{1}"
+        )
+        headers = response.response.headers
+        if "Content-Range" not in headers:
+            raise ValueError("Content-Range header not found in response headers")
+        blob_size = int(headers["Content-Range"].split("/")[1])
+        if blob_size <= 0:
+            raise ValueError("Blob size is not valid")
```
kyleknap, line 277: In terms of approaching this, there a couple of things we should keep in mind:
1. Any SDK specific logic should be kept to the `AzStorageTorchBlobClient` as the purpose of that class is to simplify our usage of the SDK in `BlobIO` by consolidating it there. For this change, we should try to keep nearly all of it to the `AzStorageTorchBlobClient` class.
2. We do not necessarily want to be replacing the get blob properties call with a single byte download as the purpose of the change is that if we are starting from a freshly opened `BlobIO` and the blob is small, we want to be able to return the contents with only a single HTTP request made when calling from `BlobIO.read()`. We are going to have to figure out how to effectively pull these headers on our very first `AzStorageTorchBlobClient.download()` call as part of its GET blob request.

src/azstoragetorch/io.py
```diff
@@ -258,7 +265,28 @@ def _should_download_more_for_readline(self, consumed: bytes, limit: int) -> boo
...
+        self._blob_properties_initialized = True
+        return blob_size
+    
     def _read(self, size: Optional[int]) -> bytes:
+        _LOGGER.debug("_read")
+        if not self._blob_properties_initialized:
+            _LOGGER.debug("Blob properties not set, setting from header")
+            self._set_blob_size_from_header()
         if size == 0 or self._is_at_end_of_blob():
             return b""
```
kyleknap, line 291: For the changes to `BlobIO`, I think the logic involving these two lines is where we will want to be targeting. Specifically. it is problematic that we are calling `self._is_at_end_of_blob()` for every read because `self._client.get_blob_size()` will may call the get blob properties API if it does not have the size cached.

Instead, we should probably explore adding logic that avoids us making a call to `client.get_blob_size()` on our very first read. To do that, I think we will need to:
1. Cache the blob size as part of the `BlobIO` object. We would initialize the value to `None` to indicate that we have not reached out to the client to get the size. The `None` value will effectively be a sentinel so that we can check if the `_blob_size` is `None` then possibly do not proceed forward with `client.get_blob_size()`.
2. In combination with (1), use the `_position` member to determine if we are at the beginning of the file. So at that point if `self._blob_size` is `None` and `self._position` equals `0`, it is a good indication that we want to avoid trying to retrieve the size.

src/azstoragetorch/_client.py
```diff
@@ -88,6 +91,7 @@ def get_blob_size(self) -> int:
         return self._blob_properties.size
 
     def download(self, offset: int = 0, length: Optional[int] = None) -> bytes:
+        _LOGGER.debug("download")
```
kyleknap, line 94: Then for the blob client, what we will probably need to do here:
1. Refactor the `_blob_properties` to not immediately try to grab the properties if we access it. Instead at the very beginning of the `download()` method, determine if the size is cached.
2. If the size is not cached, attempt a `_download_with_retries` using the length or partition threshold (whichever is less). As part of this download we will want to parse the size and etag from the headers similar to what is currently being done in `BlobIO`.
3. Continue on with the download logic if there is more content to retrieve and account for the already retrieved amount.

kyleknap (commented):
It looks like a good start! I just had suggestions on how to adjust the approach to grabbing the length and etags.

src/azstoragetorch/_client.py
```diff
@@ -185,17 +219,32 @@ def _download_with_retries(self, pos: int, length: int) -> bytes:
...
         try:
-            return self._generated_sdk_storage_client.blob.download(
+            response = self._generated_sdk_storage_client.blob.download(
                 range=f"bytes={pos}-{pos + length - 1}",
-                modified_access_conditions=azure.storage.blob._generated.models.ModifiedAccessConditions(
+            ) 
+            if self._blob_size is None: 
+                headers = response.response.headers
+                if "Content-Range" not in headers:
+                    raise ValueError("Content-Range header not found in response headers")
```
kyleknap, line 228: I don't think we need to explicitly raise if the header is not present. We always expect it to be present and so we should go ahead with retrieving the header name and if it is not present there will be a propagated on our behalf from the dictionary anyway.

In general, we should avoid raising validation errors on responses as if we are not careful, the validation can backfire and cause the library to throw errors when we were getting successful, valid responses.

src/azstoragetorch/_client.py
```diff
@@ -185,17 +219,32 @@ def _download_with_retries(self, pos: int, length: int) -> bytes:
...
     def _get_download_stream(self, pos: int, length: int) -> Iterator[bytes]:
         try:
-            return self._generated_sdk_storage_client.blob.download(
+            response = self._generated_sdk_storage_client.blob.download(
                 range=f"bytes={pos}-{pos + length - 1}",
-                modified_access_conditions=azure.storage.blob._generated.models.ModifiedAccessConditions(
+            ) 
+            if self._blob_size is None: 
+                headers = response.response.headers
+                if "Content-Range" not in headers:
```
kyleknap, line 227: It might be worth isolating this logic into a separate helper method `_set_blob_properties_from_download()` to reduce the length of the `_get_download_stream()` method.

src/azstoragetorch/_client.py
```diff
@@ -166,9 +182,27 @@ def _get_partitions(
...
     def _download_with_retries(self, pos: int, length: int) -> bytes:
         attempt = 0
-        while self._attempts_remaining(attempt):
+        while self._attempts_remaining(attempt):   
             stream = self._get_download_stream(pos, length)
+            if stream == b"":
+                return b""
             try:
+                if self._blob_size >= self._PARTITIONED_DOWNLOAD_THRESHOLD and length >= self._PARTITIONED_DOWNLOAD_THRESHOLD:
+                    streams_list = [stream]
```
kyleknap, line 191: Let's avoid making any changes to the `_download_with_retries()` method mainly it's purpose is to just do a single GET request with retries. This logic should probably go in a layer above this more part of the body of the public `download()` method.

And I think to approach this, we will want the logic structured in `download()` to be something like in terms of how it would be structured:
```python
def download(self, offset: int = 0, length: Optional[int] = None) -> bytes:
     initial_content = b""
     if self._blob_properties is None:
          initial_content = self._download_from_unknown_blob_size(offset, length)
     if self._more_to_download(initial_content, offset, length):  # Determine if we need to keep downloading here
         # This method would have the logic of what is currently in `download()` but also recalculate
         # new offsets and length to use based on pre downloaded content.
         return self._download_from_known_blob_size(offset, length, initial_content) 
     return initial_content

**Review by kyleknap — COMMENTED** [review]
It is looking better! I mainly focused on the `_client.py` part this time around since the bulk of the logic is there. We can focus on getting `_client.py` in a polished spot and then focus on polishing the other files/tests.

File: src/azstoragetorch/_client.py [hunk 7]
```diff
@@ -185,17 +219,32 @@ def _download_with_retries(self, pos: int, length: int) -> bytes:
...
+                )
+                self._blob_size = blob_size
+            return response
         except azure.core.exceptions.HttpResponseError as e:
             # TODO: This is so that we properly map exceptions from the generated client to the correct
             # exception class and error code. In the future, prior to a GA, we should consider pulling
             # in this function or a derivative of it if we plan to continue to raise Azure Python SDK
             # exceptions from this library (i.e. instead of raising our own exception classes).
+            if "InvalidRange" in str(e):
+                return b""
```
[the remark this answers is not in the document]
anjaliratnam-msft: For the test_random_seeks_and_reads when the offset is greater than the blob size it returns an error when it's expecting an empty string.

kyleknap (commented):
It's looking good. I still kept the focus of the review on `_client.py` so that we get that in a stable spot before looking at the other parts of the update.

kyleknap (commented):
It's looking good! I think after this iteration, we should be in a good spot for the implementation. Afterwords, I plan to do a pass on the tests.

kyleknap (commented):
Just had some more comments before moving onto the tests. Overall it is looking good.

tests/unit/test_client.py
```diff
@@ -50,16 +51,18 @@ def blob_properties(blob_length, blob_etag):
 
 
 @pytest.fixture
-def mock_generated_sdk_storage_client():
+def mock_generated_sdk_storage_client(blob_length):
```
kyleknap, line 54: We can probably remove `blob_length` from being used in this fixture.

tests/unit/test_client.py
```diff
@@ -50,16 +51,18 @@ def blob_properties(blob_length, blob_etag):
...
     return mock_generated_sdk_client
 
 
 @pytest.fixture
-def mock_sdk_blob_client(mock_generated_sdk_storage_client):
+def mock_sdk_blob_client(mock_generated_sdk_storage_client, blob_length):
     mock_sdk_client = mock.Mock(BlobClient)
     mock_sdk_client._client = mock_generated_sdk_storage_client
+    mock_sdk_client._client._blob_properties = mock.Mock()
+    mock_sdk_client._client._blob_properties.size = blob_length
```
kyleknap, line 65: I'm a little hesitant to be accessing even more internal properties as it generally will make it harder to do refactorings in the future. Could we just set the `return_value` for the `get_blob_properties` method? I think that could make sense especially since most tests are setting the return value for that method.

tests/unit/test_client.py
```diff
@@ -104,6 +107,23 @@ def to_bytes_iterator(content, chunk_size=64 * 1024, exception_to_raise=None):
...
+    response.response.headers = {
+        "Content-Range": f"bytes {expected_range}/{blob_length}",
+        "ETag": etag,
+    }
+    response.iter_content_func = iter(
+        to_bytes_iterator(
+            content=slice_bytes(content, expected_range), exception_to_raise=exception
+        )
+    )
+    response.__iter__ = mock.Mock()
```
kyleknap, line 122: Instead of setting a `Mock()` object like this, we can just make the `response` a [`MagicMock`](https://docs.python.org/3/library/unittest.mock.html#magicmock-and-magic-method-support) which will allow us to mock the `__iter__` return value.

tests/unit/test_client.py
```diff
@@ -104,6 +107,23 @@ def to_bytes_iterator(content, chunk_size=64 * 1024, exception_to_raise=None):
...
 
 
+def mock_response(expected_range, blob_length, content, exception=None, etag=blob_etag):
+    response = mock.Mock(spec=StreamDownloadGenerator)
+    response.response = mock.Mock()
+    response.response.headers = {
+        "Content-Range": f"bytes {expected_range}/{blob_length}",
+        "ETag": etag,
+    }
+    response.iter_content_func = iter(
```
kyleknap, line 117: Is there a particular reason we need to wrap the iterator from `to_bytes_iterator` with another iterator from `iter()`?

tests/unit/test_client.py
```diff
@@ -327,25 +360,42 @@ def test_close(self, mock_sdk_blob_client):
...
         expected_ranges,
+        known_blob_size,
         azstoragetorch_blob_client,
         mock_sdk_blob_client,
         mock_generated_sdk_storage_client,
         blob_properties,
     ):
         blob_properties.size = blob_size
         mock_sdk_blob_client.get_blob_properties.return_value = blob_properties
+        if not known_blob_size:
```
kyleknap, line 378: I like that we tried reusing the existing cases without changing them to add these new cases, but looking more through how the test is setup, I think we should just make explicit test cases for `known_blob_size` and explicit expected ranges. Specifically, we'd:
1. Hoist `known_blob_size` into the main `parameterize` decorator.
2. Update all of the current test cases with a `True` to keep their current test cases
3. Add a case directly below the `True` cases with the appropriate expected ranges.

I realize this is going to be a little tedious, but I think it is more of a reflection of the complexity of what we are adding to the `download()` method.

In general, I'm not sure if we should adding logic programmatic logic to generate what is expected because we could have the same bug in the programmatic logic as the actual logic. By having the expected ranges hard coded when the blob size is not known, it is explicit in what we expect which makes it easier to understand what we are expecting and also less of a chance of our test cases being wrong.

tests/unit/test_client.py
```diff
@@ -104,6 +107,23 @@ def to_bytes_iterator(content, chunk_size=64 * 1024, exception_to_raise=None):
             raise exception_to_raise
 
 
+def mock_response(expected_range, blob_length, content, exception=None, etag=blob_etag):
```
kyleknap, line 110: For the signature, let's:

1. Update the name to `mock_download_response()` to make it a little more clear what type of response we are mocking
2. Remove the `blob_etag` default as that is a fixture and that cannot be used when this function is not a fixture. Instead we can just set the `etag` default to `None`.

tests/unit/test_client.py
```diff
@@ -136,17 +156,30 @@ def submit(self, fn, *args, **kwargs):
...
+        expected_download_calls = []
+        start = 0
+        if not known_blob_size:
+            expected_download_calls.append(
+                mock.call(
+                    range=f"bytes={expected_ranges[0]}",
+                )
+            )
+            start = 1
+        for i in range(start, len(expected_ranges)):
```
kyleknap, line 174: To simplify this down, we can use [`enumerate()`](https://docs.python.org/3/library/functions.html#enumerate) in the for loop and check for `i == 0 and not known_blob_size` to whether we inject the modified access condition as part of the expected call

kyleknap (commented):
Nice! The implementation looks like it is in a good spot and thought we are in a fairly good spot on the tests as well. I mainly had suggestions related to refactoring the tests a bit more.

tests/unit/test_io.py
```diff
@@ -286,6 +286,10 @@ def test_read(self, blob_io, blob_content, mock_azstoragetorch_blob_client):
         mock_azstoragetorch_blob_client.download.assert_called_once_with(
             offset=0, length=None
         )
+        calls = mock_azstoragetorch_blob_client.mock_calls
+        assert calls[0] == mock.call.download(offset=0, length=None)
+        assert calls[1] == mock.call.get_blob_size()
+        mock_azstoragetorch_blob_client.get_blob_size.assert_called_once()
```
kyleknap, line 292: For this part, we should be able to reduce this down to:
```python
assert mock_azstoragetorch_blob_client.mock_calls == [
    mock.call.download(offset=0, length=None),
    mock.call.get_blob_size()
]
```

kyleknap (commented):
Looks good! Just had a few more comments on the tests but should be set here.

tests/unit/test_client.py
```diff
@@ -673,6 +673,9 @@ def test_no_executor_used_when_no_transfers(self, mock_sdk_blob_client):
     @pytest.mark.parametrize(
         "blob_size, download_offset, download_length, expected_ranges, known_blob_size",
         [
+            # Download an empty blob
+            (0, None, None, ["0--1"], True),
+            (0, None, None, [f"0-{DEFAULT_PARTITION_SIZE - 1}"], False),
```
kyleknap, line 678: It would also be good to add a cases where the invalid range exception is thrown where:
1. We still raise the exception if the size is larger than 0 bytes
2. We return empty bytes if the size is 0 bytes

I think we should be able to add a new test method similar to how `test_maps_download_exceptions` is set up.

kyleknap (commented):
Looks good. Just had one suggestion on the tests.

kyleknap (commented):
Looks good. Just had these last few comments on the tests and we should be good to merge.

PR #29, opened 2025-04-18
Add support for map-style datasets

kyleknap:
It introduces the `datasets` module with its first dataset class, `BlobDataset`. It is a [map-style dataset](https://pytorch.org/docs/stable/data.html#map-style-datasets) intended for random sampling of dataset loaded into memory. The dataset class itself allows for accessing blobs listed from a container (e.g., `from_container_url`) or a customer provided list of blobs (e.g., `from_blob_urls`). In addition, customers can provide a custom transform that accepts a `Blob` object that can be used to access blob content and identifiers as part of the transform.

One additional refactor that needed to be made was to make executor instantiation in the blob client lazy. The executor itself is not pickleable when the PyTorch data loader uses workers. By making it lazy, it ensures that it is created once the process have been created and start reading data.

Note this PR does not include the following, which will be sent as subsequent PRs to help break up the review:
* [Iterable-style datasets](https://pytorch.org/docs/stable/data.html#iterable-style-datasets)
* End-to-end tests
* Docstrings for new public interfaces

### DIFF
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -15,7 +15,7 @@ import threading
 import time
 import urllib.parse
 import uuid
-from typing import Optional, List, Tuple, Iterator, Union, Literal
+from typing import Optional, List, Tuple, Iterator, Union, Literal, TypedDict
 
 from azure.core.credentials import (
     AzureSasCredential,
@@ -44,6 +44,13 @@ SUPPORTED_WRITE_BYTES_LIKE_TYPE = Union[bytes, bytearray, memoryview]
 STAGE_BLOCK_FUTURE_TYPE = concurrent.futures.Future[str]
 
 
+class SDKKwargsType(TypedDict, total=False):
+    connection_data_block_size: int
+    transport: RequestsTransport
+    user_agent: str
+    credential: SDK_CREDENTIAL_TYPE
+
+
 class AzStorageTorchBlobClientFactory:
     # Socket timeouts set to match the default timeouts in Python SDK
     _SOCKET_CONNECTION_TIMEOUT = 20
@@ -60,6 +67,17 @@ class AzStorageTorchBlobClientFactory:
     def get_blob_client_from_url(self, blob_url: str) -> "AzStorageTorchBlobClient":
         blob_sdk_client = self._get_sdk_blob_client_from_url(blob_url)
         return AzStorageTorchBlobClient(blob_sdk_client)
+
+    def yield_blob_clients_from_container_url(
+        self, container_url: str, prefix: Optional[str] = None
+    ) -> Iterator["AzStorageTorchBlobClient"]:
+        container_sdk_client = self._get_sdk_container_client_from_container_url(
+            container_url
+        )
+        blob_names = container_sdk_client.list_blob_names(name_starts_with=prefix)
+        for blob_name in blob_names:
+            blob_client = container_sdk_client.get_blob_client(blob_name)
+            yield AzStorageTorchBlobClient(blob_client)
 
     def _get_sdk_credential(
         self, credential: AZSTORAGETORCH_CREDENTIAL_TYPE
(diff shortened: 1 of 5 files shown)


src/azstoragetorch/_client.py
```diff
@@ -182,6 +237,18 @@ def _get_max_in_flight_requests(self) -> int:
         cpu_count_fn = getattr(os, "process_cpu_count", os.cpu_count)
         return min(32, (cpu_count_fn() or 1) + 4)
 
+    def _get_executor(self) -> concurrent.futures.Executor:
+        # We want executor creation to be lazy instead of instantiating immediately in
+        # the constructor because the executor itself is not pickleable. This is an issue
+        # when workers are used by PyTorch's DataLoader as workers are spawned as processes.
+        # So we delay executor creation until it is needed for reading/writing data which will
+        # happen after processes are spawned.
+        if self._executor is None:
```
jalauzon-msft, line 246: Any issues with concurrency here? I assume only a single thread will ever be calling this for each class?

src/azstoragetorch/_client.py
```diff
@@ -282,3 +351,34 @@ def _stage_block(self, data: SUPPORTED_WRITE_BYTES_LIKE_TYPE) -> str:
...
     def _release_in_flight_semaphore(self, _: STAGE_BLOCK_FUTURE_TYPE) -> None:
         self._max_in_flight_semaphore.release()
+
+    def _get_url_with_filtered_query_string(
+        self, parsed_url: urllib.parse.ParseResult
+    ) -> str:
+        # Helper method to only include indentifying query string parameters for a blob URL. More specifically,
+        # we do not want to return any SAS tokens in the URL as it can accidentally result in leaking credentials
+        # as part of interfaces that expose the URL (e.g., azstoragetorch.datasets.Blob).
+        return urllib.parse.urlunparse(
```
jalauzon-msft, line 361: As discussed offline, let's just filter out all query params for now.

src/azstoragetorch/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+    ):
+        self._blobs = blobs
+        self._transform = transform
+
+    @classmethod
+    def from_blob_urls(
+        cls,
+        blob_urls: Union[str, Iterable[str]],
+        *,
+        credential: _client.AZSTORAGETORCH_CREDENTIAL_TYPE = None,
```
jalauzon-msft, line 69: I'm not sure if we've done this in other places but this means we are defaulting to `DefaultAzureCredential`? I'm not sure if that is the best choice or not? Would defaulting to `False` (no credential) be better? The SDK always defaults to no credential but I'm not sure if we want to follow suit here or not.

src/azstoragetorch/_client.py
kyleknap: There should not be an issue based on how the client is used today. Specifically, these methods should only be called from a single thread because the public interfaces that use the method (e.g. `BlobIO`) should only ever be handled by a single thread. If customers are using multiple threads (which is not an expected workflow), they will need to make sure to be protecting access to the `BlobIO` with a lock.

src/azstoragetorch/datasets.py
kyleknap: We do it today in `BlobIO`. That was chosen because:

1. It leans into more of a "batteries-included" approach for the project. In most cases, customers will want authorization. If we were to have authorization be off by default, it would mean that most customers will need to specify credentials for any `BlobIO` object, making the call patterns more verbose (e.g., `BlobIO("url", "rb", credential=DefaultAzureCredential()`), which also makes any compatibility shims with `open()` more complicated.
2. By defaulting to an Oauth approach, we are defaulting to the recommend way to do authorization in Azure today and general would have a more secure default than not adding credentials.
3. By using `DefaultAzureCredential` it casts the widest coverage for making authentication just work for the library no matter the environment. It also means customers are spending less time needing to figure out credentials to get up and running especially if they are not too familiar with SDK concepts.

The datasets approach to credentials was chosen to be consistent with `BlobIO` and for the reasons above as well.

jalauzon-msft: Ok, sounds good.

jalauzon-msft (approved):

PR #31, opened 2025-04-22
Implement iterable-style dataset

kyleknap:
Includes built-in support for worker sharding. Also refactored the datasets to share how blob client factories and blob objects are
returned by consolidating the logic into BlobIterables.

The PR also includes a commit to make semaphore creation of the blob client lazy. This logic was in a branch where I have e2e tests staged (and will be submitted as a later PR) but noticed it was not upstreamed to the final dataset implementation. So, I added it as part of this PR. Both datasets pass multiple-worker PyTorch DataLoader tests now. I also double checked there was nothing else missing from the e2e test branch; when the PR is submitted for e2e tests, it should be only test related changes.

### DIFF
--- a/src/azstoragetorch/datasets.py
+++ b/src/azstoragetorch/datasets.py
@@ -4,7 +4,7 @@
 # license information.
 # --------------------------------------------------------------------------
 
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Iterator
 from typing import Optional, Union, TypeVar, TypedDict
 
 import torch.utils.data
@@ -56,9 +56,9 @@ class Blob:
 
 class BlobDataset(torch.utils.data.Dataset[_TransformOutputType_co]):
     def __init__(
-        self, blobs: list[Blob], transform: _TRANSFORM_TYPE = _default_transform
+        self, blobs: Iterable[Blob], transform: _TRANSFORM_TYPE = _default_transform
     ):
-        self._blobs = blobs
+        self._blobs = list(blobs)
         self._transform = transform
 
     @classmethod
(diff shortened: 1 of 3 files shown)


src/azstoragetorch/datasets.py
```diff
@@ -106,3 +92,93 @@ def __getitem__(self, index: int) -> _TransformOutputType_co:
...
+        self._transform = transform
+
+    @classmethod
+    def from_blob_urls(
+        cls,
+        blob_urls: Union[str, Iterable[str]],
+        *,
+        credential: _client.AZSTORAGETORCH_CREDENTIAL_TYPE = None,
+        transform: _TRANSFORM_TYPE = _default_transform,
+    ) -> "IterableBlobDataset":
```
jalauzon-msft, line 111: Sorry I didn't catch this before in other places but is `typing.Self` a better return type here? I think `typing.Self` does a better job with inheritance typing. [PEP](https://peps.python.org/pep-0673/)

It is Python 3.11 though so we either would need `typing_extensions` and/or a conditional import.

jalauzon-msft (approved):

src/azstoragetorch/datasets.py
kyleknap: That's a good callout. I do think `typing.Self` is better a return type as well. I also think it handles generics too, which I think got missed in this return type (e.g. `IterableBlobDataset[_TransformOutputType_co]`)

In terms of adding `typing_extensions`, that sounds good to me. The dependency is already in the library's tree since the SDKs use it. I've also seen other instances where it would have been nice to be able to use some typing backports.

I'm going to merge this PR as is and then send a follow up PR that adds `typing_extensions` to the project and do a pass of updating the types for datasets.

PR #33, opened 2025-04-23
Fill out documentation intro and add user guide

kyleknap:
For the user guide, it includes information on getting started and information on how to save and load models using BlobIO.

### DIFF
--- a/doc/sphinx/index.rst
+++ b/doc/sphinx/index.rst
@@ -1,7 +1,22 @@
 Azure Storage for PyTorch Documentation
 =======================================
 
-Azure Storage integrations for PyTorch.
+Azure Storage for PyTorch (``azstoragetorch``) is a library that provides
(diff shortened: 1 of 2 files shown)


doc/sphinx/index.rst
```diff
@@ -1,11 +1,29 @@
...
+
 
 API Reference
 -------------
 .. toctree::
    :maxdepth: 2
 
    api
+
+.. _Azure Storage: https://learn.microsoft.com/en-us/azure/storage/common/storage-introduction
```
jalauzon-msft, line 28: For all learn links, we should manually remove the `en-us` from the URL so that it will direct to user's default locale.

jalauzon-msft (approved):

doc/sphinx/index.rst
kyleknap: Good catch. I copied these URLs post redirection it seems. I'll update before merging.

PR #34, opened 2025-04-24
Improve typing for datasets

kyleknap:
Transform type is now properly plumbed through so based on the transform that is used the correct type is propagated. This was done by:

1. Introducing a `default` of the default typed dict to our transform type var
2. Use `Self` for the return type to ensure we include the generic as part of the type
3. Move the default transform out of the signature and instead use `None` so `mypy` would not always assume a string was the generic
4. Stop using a shared `Callable` alias and instead expand the callable expression for each `transform` argument. I was finding `mypy` would resolve the shared callable to the default type and apply that resolution to all uses of it.

This also required adding `typing-extensions`. The library already indirectly depends on `typing-extensions` so it is not a net new dependency.

To test that it is working as expected, I wrote a script with purposeful wrong type usage to make sure mypy properly flagged incorrect usage and did not flag correct usage:
```python
from azstoragetorch.datasets import BlobDataset, Blob

BLOB_URL = "https://<account>.blob.core.windows.net/<container>/<blob>"

dataset = BlobDataset.from_blob_urls(BLOB_URL)

# These should pass mypy for default transform
dataset[0]["url"]
dataset[0]["data"]

# These should run into mypy issues
dataset[0].url  # Dicts do not have url method
dataset[0]["not-a-key"]  # Access a key that does not exist


def to_url(blob: Blob) -> str:
    return blob.url


dataset_with_transform = BlobDataset.from_blob_urls(BLOB_URL, transform=to_url)

# Type should be str so we should be able to do things like concat strings
dataset_with_transform[0] + "add string together"

# These should run into mypy issues
dataset_with_transform[0]["url"]  # Complain about slicing with a string
dataset_with_transform[0]["data"]
dataset_with_transform[0].not_a_method()  # String does not have this method


# This will fail because transform does not match specified type of int
dataset_with_wrong_transform: BlobDataset[int] = BlobDataset.from_blob_urls(
    BLOB_URL, transform=to_url
)
```
Running `mpy` on it results in the expected errors:
```
> mypy .\check-dataset-type.py
check-dataset-type.py:12: error: "_DefaultTransformOutput" has no attribute "url"  [attr-defined]
check-dataset-type.py:13: error: TypedDict "_DefaultTransformOutput" has no key "not-a-key"  [typeddict-item]
check-dataset-type.py:26: error: Invalid index type "str" for "str"; expected type "SupportsIndex | slice[Any, Any, Any]"  [index]
check-dataset-type.py:27: error: Invalid index type "str" for "str"; expected type "SupportsIndex | slice[Any, Any, Any]"  [index]
check-dataset-type.py:28: error: "str" has no attribute "not_a_method"  [attr-defined]
check-dataset-type.py:33: error: Argument "transform" to "from_blob_urls" of "BlobDataset" has incompatible type "Callable[[Blob], str]"; expected "Callable[[Blob], int] | None"  [arg-type]
Found 6 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/src/azstoragetorch/datasets.py
+++ b/src/azstoragetorch/datasets.py
@@ -5,7 +5,8 @@
 # --------------------------------------------------------------------------
 
 from collections.abc import Callable, Iterable, Iterator
-from typing import Optional, Union, TypeVar, TypedDict
+from typing import Optional, Union, TypedDict
+from typing_extensions import Self, TypeVar
 
 import torch.utils.data
 
@@ -13,8 +14,9 @@ from azstoragetorch.io import BlobIO
 from azstoragetorch import _client
 
 
-_TransformOutputType_co = TypeVar("_TransformOutputType_co", covariant=True)
-_TRANSFORM_TYPE = Callable[["Blob"], _TransformOutputType_co]
+_TransformOutputType_co = TypeVar(
+    "_TransformOutputType_co", covariant=True, default="_DefaultTransformOutput"
+)
 
 
 class _DefaultTransformOutput(TypedDict):
@@ -56,9 +58,13 @@ class Blob:
 
 class BlobDataset(torch.utils.data.Dataset[_TransformOutputType_co]):
     def __init__(
-        self, blobs: Iterable[Blob], transform: _TRANSFORM_TYPE = _default_transform
+        self,
+        blobs: Iterable[Blob],
+        transform: Optional[Callable[[Blob], _TransformOutputType_co]] = None,
     ):
         self._blobs = list(blobs)
+        if transform is None:
+            transform = _default_transform
         self._transform = transform
 
     @classmethod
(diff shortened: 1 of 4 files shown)


src/azstoragetorch/datasets.py
```diff
@@ -5,16 +5,18 @@
 # --------------------------------------------------------------------------
 
 from collections.abc import Callable, Iterable, Iterator
-from typing import Optional, Union, TypeVar, TypedDict
+from typing import Optional, Union, TypedDict
+from typing_extensions import Self, TypeVar
```
jalauzon-msft, line 9: With `typing_extensions` imports what we've done in the past is conditional imports based on Python version. So something like:

```py
if sys.version_info >= (3, 11):
    from typing import Self
else:
    from typing_extensions import Self
```

This more easily shows when we can remove the use of typing_extensions for that import. I'll leave it up to you though if we want to do that here.

jalauzon-msft (approved):

src/azstoragetorch/datasets.py
kyleknap: I'm not sure if I have too strong of an opinion.

Having the conditional imports is nice especially if we get to the point where we no longer need `typing_extensions` we can consider even dropping the dependency, but doubtful we would be able to prune it completely from the dependency tree as a dependency will likely use it.

Not having the conditional import can also be nice because it guarantees we are using the same implementation no matter the Python version and there is less version branching.

I think I'm going to keep it as is. If we were to start doing it, I'd prefer we have a `_compat.py` file that houses all of the conditional compatibility logic and I'm hesitant to do it quite yet because there is only one other compatibility-related line in the calculating default executor workers. If there is another opportunity that comes up in the library, I can look into adding it.

PR #35, opened 2025-04-24
Add API references for datasets module

kyleknap:
Uses Sphinx autodoc for the `BlobDataset`, `BlobIterableDataset`, and `Blob` classes

### DIFF
--- a/src/azstoragetorch/datasets.py
+++ b/src/azstoragetorch/datasets.py
@@ -35,28 +35,101 @@ def _default_transform(blob: "Blob") -> _DefaultTransformOutput:
 
 
 class Blob:
+    """Object representing a single blob in a dataset.
+
+    Datasets instantiate :py:class:`Blob` objects and pass them directly to a dataset's
+    ``transform`` callable. Within the ``transform`` callable, use properties and methods
+    to access a blob's properties and content. For example::
+
+        from azstoragetorch.datasets import Blob, BlobDataset
+
+        def to_bytes(blob: Blob) -> bytes:
+            with blob.reader() as f:
+                return f.read()
+
+        dataset = BlobDataset.from_blob_urls(
+            "https://<storage-account-name>.blob.core.windows.net/<container-name>/<blob-name>",
+            transform=to_bytes
+        )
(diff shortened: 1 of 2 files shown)


src/azstoragetorch/datasets.py
```diff
@@ -35,28 +35,101 @@ def _default_transform(blob: "Blob") -> _DefaultTransformOutput:
...
+
+        def to_bytes(blob: Blob) -> bytes:
+            with blob.reader() as f:
+                return f.read()
+
+        dataset = BlobDataset.from_blob_urls(
+            "https://<storage-account-name>.blob.core.windows.net/<container-name>/<blob-name>",
+            transform=to_bytes
+        )
+        print(type(dataset[0]))  # Type should be: <class 'bytes'>
```
jalauzon-msft, line 54: Is there a way to write code samples in a .py file and then "link" them into the docs? I thought we had something like this for the SDK but maybe that was an Azure SDK specific thing. If it is possible, maybe we should consider that for this and others that came in previous PRs. Just seems better to have the code in a .py file so it's less easy to make a mistake (unless IDEs are smart enough to help out in the doc too).

Definitely something that we could do in a later PR if we decide to.

jalauzon-msft (approved):

src/azstoragetorch/datasets.py
kyleknap: I looked into how the SDK does it. It uses the Sphinx [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) directive. From what I recall, this is how I've done embedded examples in the past for documentation. So I think we'd explore using that first.

Yeah I agree it would be interesting to just having standalone samples at some point. The HTML pages have easy copy and paste buttons to use. So, from a usability perspective that should be good for the initial version of the docs.

Mainly by having the samples in a standalone file, it is easier to test and also easier to bulk manage samples. Though referencing examples can be a little brittle if we end up changing the content and throws off how Sphinx pulls in the example. So, we need to be mindful about it.

PR #37, opened 2025-04-28
Add datasets user guide

kyleknap:
The focus of the PR was to get all of the content that we want down for a user guide on datasets. I think there is some additional updates that would be good to do at some point:
* Break out samples as separate files. There were times where the samples felt like they could be shared between the guide and API reference
* Explore introducing [tabs](https://sphinx-design.readthedocs.io/en/furo-theme/tabs.html) to toggle between dataset class examples (instead of making it one large example with both dataset classes created). I did not do that because it makes the raw ReStructured text more complex and would be easier to manage if we had the code in samples.

### DIFF
--- a/doc/sphinx/user-guide.rst
+++ b/doc/sphinx/user-guide.rst
@@ -93,6 +93,223 @@ specify the URL to the blob storing the model weights and use read mode (i.e., `
         model.load_state_dict(torch.load(f))
 
 
+.. _datasets-guide:
+
+PyTorch Datasets
+----------------
+
+PyTorch offers the `Dataset and DataLoader primitives <PyTorch dataset tutorial_>`_ for
+loading data samples. ``azstoragetorch`` provides implementations for both types
+of PyTorch datasets, `map-style and iterable-style datasets <PyTorch dataset types_>`_,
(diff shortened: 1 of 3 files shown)


doc/sphinx/user-guide.rst
```diff
@@ -93,10 +93,233 @@ specify the URL to the blob storing the model weights and use read mode (i.e., `
...
+    import torchvision.transforms  # Install separately: ``pip install torchvision``
+
+    # Update URL with your own Azure Storage account, container, and blob containing an image
+    IMAGE_BLOB_URL = "https://<storage-account-name>.blob.core.windows.net/<container-name>/<blob-image-name>"
+
+    # Define transform to convert blob to a tuple of (image_name, image_tensor)
+    def to_img_name_and_tensor(blob: Blob) -> tuple[str, torch.Tensor]:
+        # Use blob reader to retrieve blob contents and then transform to an image tensor.
+        with blob.reader() as f:
+            image = PIL.Image.open(f)
```
jalauzon-msft, line 238: Does this thing need to be closed or used in context manager?

jalauzon-msft (approved):

PR #46, opened 2025-05-08
Modified sphinx docs and added samples

anjaliratnam-msft:
Added sphinx-tabs for the different dataset types in the docs and added samples as well to reuse examples in the docs.

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -28,6 +28,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.intersphinx",
     "sphinx_copybutton",
+    "sphinx_design",
 ]
 
 intersphinx_mapping = {
--- a/doc/sphinx/user-guide.rst
+++ b/doc/sphinx/user-guide.rst
@@ -53,44 +53,20 @@ Saving a Model
 ~~~~~~~~~~~~~~
 To save a model to Azure Blob Storage, pass a :py:class:`azstoragetorch.io.BlobIO`
 directly to :py:func:`torch.save`. When creating the :py:class:`~azstoragetorch.io.BlobIO`,
-specify the URL to the blob you'd like to save the model to and use write mode (i.e., ``wb``)::
+specify the URL to the blob you'd like to save the model to and use write mode (i.e., ``wb``)
 
-    import torch
-    import torchvision.models  # Install separately: ``pip install torchvision``
-    from azstoragetorch.io import BlobIO
-
-    # Update URL with your own Azure Storage account and container name
-    CONTAINER_URL = "https://<my-storage-account-name>.blob.core.windows.net/<my-container-name>"
-
-    # Model to save. Replace with your own model.
-    model = torchvision.models.resnet18(weights="DEFAULT")
-
-    # Save trained model to Azure Blob Storage. This saves the model weights
-    # to a blob named "model_weights.pth" in the container specified by CONTAINER_URL.
-    with BlobIO(f"{CONTAINER_URL}/model_weights.pth", "wb") as f:
-        torch.save(model.state_dict(), f)
+    .. literalinclude:: ../../samples/save_model.py
+        :lines: 9-
 
 
 Loading a Model
 ~~~~~~~~~~~~~~~
 To load a model from Azure Blob Storage, pass a :py:class:`azstoragetorch.io.BlobIO`
 directly to :py:func:`torch.load`. When creating the :py:class:`~azstoragetorch.io.BlobIO`,
-specify the URL to the blob storing the model weights and use read mode (i.e., ``rb``)::
-
-    import torch
-    import torchvision.models  # Install separately: ``pip install torchvision``
-    from azstoragetorch.io import BlobIO
-
-    # Update URL with your own Azure Storage account and container name
-    CONTAINER_URL = "https://<my-storage-account-name>.blob.core.windows.net/<my-container-name>"
-
-    # Model to load weights for. Replace with your own model.
-    model = torchvision.models.resnet18()
-
-    # Load trained model from Azure Blob Storage.  This loads the model weights
-    # from the blob named "model_weights.pth" in the container specified by CONTAINER_URL.
-    with BlobIO(f"{CONTAINER_URL}/model_weights.pth", "rb") as f:
-        model.load_state_dict(torch.load(f))
+specify the URL to the blob storing the model weights and use read mode (i.e., ``rb``)
+    
+    .. literalinclude:: ../../samples/load_model.py
+        :lines: 9-
 
 
 .. _datasets-guide:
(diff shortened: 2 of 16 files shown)


doc/sphinx/user-guide.rst
```diff
@@ -170,26 +143,16 @@ corresponding ``from_blob_urls()`` method:
...
-    # Create a map-style dataset from the list of blob URLs
-    map_dataset = BlobDataset.from_blob_urls(blob_urls)
+The method accepts a list of blob URLs to create the dataset from. For example
 
-    # Create an iterable-style dataset from the list of blob URLs
-    iterable_dataset = IterableBlobDataset.from_blob_urls(blob_urls)
+.. tabs::
+    .. group-tab:: BlobDataset
+        .. literalinclude:: ../../samples/blob_dataset_sample.py
+            :lines: 16, 20, 62-75
```
kyleknap, line 151: So instead of using `lines` directly, could we use the pattern the Python SDK [uses](https://github.com/Azure/azure-sdk-for-python/blob/62ea31c6c8a9afdf993a77b263d4b1a4caa10086/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L149) with the `start-after` and `end-before` properties to designate where samples begin and end? Mainly I'm concerned that the line numbers could easily be thrown out of sync if the samples get updated.  In terms of getting the imports to be included as well, I'm open to ideas, but I think we are going to have to experiment with it.

In general, it may be worth making a pass through the [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) documentation to see what else we can leverage. Specifically, it looks like it can also handle dedenting which I think we will want as well to fix the indentation issues.

doc/sphinx/conf.py
```diff
@@ -28,6 +28,7 @@
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
     'sphinx_copybutton',
+    'sphinx_tabs.tabs',
```
kyleknap, line 31: Instead of the `sphinx-tabs` library, let's use [`sphinx-design`](https://github.com/executablebooks/sphinx-design). Mainly the tabs look better and they are specifically designed to work with our theme, `furo`. It should not be fairly straightforward to convert. I was able to update one of the samples below to the new extension format and overall looked much cleaner in the rendered HTML:
```rst
.. tab-set::

    .. tab-item:: ``BlobDataset``
    
        .. literalinclude:: ../../samples/blob_dataset_sample.py
            :lines: 16, 20, 39-45
    
    .. tab-item:: ``IterableBlobDataset``
    
        .. literalinclude:: ../../samples/iterable_blob_dataset_sample.py 
            :lines: 16, 20, 39-45
```

samples/saving_and_loading_model_sample.py
```diff
@@ -0,0 +1,66 @@
...
+
+if __name__ == "__main__":
+    my_storage_account_name = os.environ["AZSTORAGETORCH_STORAGE_ACCOUNT_NAME"]
+    my_container_name = "model-sample-container"
+    setup(my_storage_account_name, my_container_name)
+    
+    # Update URL with your own Azure Storage account and container name
+    CONTAINER_URL = (
+        f"https://{my_storage_account_name}.blob.core.windows.net/{my_container_name}"
+    )
```
kyleknap, line 63: Instead of defining the `CONTAINER_URL` down here, let's define it near the top of the script below the imports as that is generally where module level constants should be defined. It will also let us remove needing to providing `CONTAINER_URL` to the `load_model()` and `save_model()` functions.

samples/saving_and_loading_model_sample.py
```diff
@@ -0,0 +1,66 @@
...
+    # Load trained model from Azure Blob Storage.  This loads the model weights
+    # from the blob named "model_weights.pth" in the container specified by CONTAINER_URL.
+    with BlobIO(f"{CONTAINER_URL}/model_weights.pth", "rb") as f:
+        model.load_state_dict(torch.load(f))
+
+
+if __name__ == "__main__":
+    my_storage_account_name = os.environ["AZSTORAGETORCH_STORAGE_ACCOUNT_NAME"]
+    my_container_name = "model-sample-container"
+    setup(my_storage_account_name, my_container_name)
```
kyleknap, line 58: So for our escape hatch to run this programatically without needing to change the sample, I'm thinking we define a function that overrides the module level `CONTAINER_URL`:
```python
def update_container_url_from_env():
    if "AZSTORAGETORCH_STORAGE_CONTAINER_URL" in os.environ:
        global CONTAINER_URL
        CONTAINER_URL = os.environ["AZSTORAGETORCH_STORAGE_CONTAINER_URL"]
```
Generally, [`global`](https://docs.python.org/3/reference/simple_stmts.html#global) we want to avoid using as it can lead to unexpected side effects but it can be useful in scenarios like this.

samples/iterable_blob_dataset_sample.py
```diff
@@ -0,0 +1,87 @@
...
+    CONTAINER_URL = (
+        f"https://{my_storage_account_name}.blob.core.windows.net/{my_container_name}"
+    )
+
+    # List of blob URLs to create dataset from. Update with your own blob names.
+    blob_urls = [
+        f"{CONTAINER_URL}/<blob-name-1>",
+        f"{CONTAINER_URL}/<blob-name-2>",
+        f"{CONTAINER_URL}/<blob-name-3>",
+    ]
```
kyleknap, line 73: For the `blob_urls`, I'm thinking we can just make this a module level constant `BLOB_URLS` and we can manage it similar to `CONTAINER_URL` in terms of using a `global` to update it from environment variables. For the environment variable name, we can just make it `AZSTORAGETORCH_BLOB_NAMES` and the value can be a comma delineated list of blob names that we can use to update the `BLOB_URLS` constant. We can also remove the logic for uploading blobs from this sample and into a `_bootstrap.py` script that we can run.

samples/iterable_blob_dataset_sample.py
```diff
@@ -0,0 +1,87 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+"""
+FILE: iterable_blob_dataset_sample.py
```
kyleknap, line 8: Instead of partitioning the dataset samples by classes, it probably makes sense to do it by class methods (e.g. `from_container_url` and `from_blob_urls`) mainly when customers come to find samples they are more likely to be searching by class method since the two differ fairly significantly in both call pattern and use case. So that will make it easier to find based on the use case a customer is looking for.

kyleknap (commented):
It's looking good. Mainly had some general suggestions to try to polish the approach and make sure we get the samples in a place that we can feel comfortable working with going forward.

samples/iterable_dataset/dataset_from_blob_list.py
```diff
@@ -0,0 +1,21 @@
+"""
+This sample demonstrates the way to create an iterable-style dataset from a blob list using the IterableBlobDataset class.
+"""
```
kyleknap, line 3: For this docstring and the other docstrings, we can probably shorten the prose even more and make it fit to a single line. For example:
```
"""Creates an iterable-style dataset from a list of blobs"""
```
As a rule of thumb, if the docstring is a one-liner, the recommendation is to include double quotes on the same line.

samples/iterable_dataset/dataset_from_container_url.py
```diff
@@ -0,0 +1,14 @@
+"""
```
kyleknap, line 1: Let's also make sure to include the copyright comment before the docstring for all of these samples.

src/azstoragetorch/io.py
```diff
@@ -20,26 +20,13 @@ class BlobIO(io.IOBase):
...
 
     Use this class directly for PyTorch checkpointing by passing it directly to
-    :py:func:`torch.save()` or :py:func:`torch.load()`::
+    :py:func:`torch.save()` or :py:func:`torch.load()`
 
-        import torch
-        import torchvision.models
-        from azstoragetorch.io import BlobIO
+        .. literalinclude:: ../../samples/save_model.py
+            :start-after: [START sample]
```
kyleknap, line 27: For this docstring it may be worth putting some non-code verbage between the two samples as when rendered, it looks like they merge together and end up getting imports in the middle of the code sample. So instead we could add before each example:
```
Sample below shows how to use ::py:class:`BlobIO` with :py:func:`torch.save()
 
# add literalinclude here

Sample below shows how to use ::py:class:`BlobIO` with :py:func:`torch.load()`

# add literalinclude here
```

samples/iterable_dataset/dataset_from_blob_list.py
```diff
@@ -0,0 +1,21 @@
...
+
+# [START sample]
```
kyleknap, line 5: If we are able to fit the docstring to one line, I'm wondering if we can just use the docstring quotes (e.g., `"""` as the sentinel value)? I think it would be worth to explore to possibly make it less awkward in having these `START` comments in all of these samples.

kyleknap (commented):
I like how we are taking this. It's definitely simplifying managing all of these samples. Just had some comments.

kyleknap (commented):
Looks good! :ship: We can merge it once GitHub Actions starts working again.

kyleknap (approved):

PR #49, opened 2025-05-12
E2E workflow updates

anjaliratnam-msft:
Modified the workflow so integration tests are run when there are pushes to main or PRs.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,50 @@
+name: E2E tests
+on:
+  push:
+    branches:
+      - e2e_workflow
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+permissions:
+  id-token: write
+  contents: read
+jobs:
+  run-ci:
+    name: Run CI
+    uses: ./.github/workflows/ci.yml
+  run-e2e-tests:
+    name: Run tests
+#    needs: build-dists
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.9", "3.10", "3.11", "3.12", "3.13" ]
+        os: [ ubuntu-latest, macOS-latest, windows-latest ]
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -58,3 +59,38 @@ jobs:
...
+  run-e2e-tests:
+    name: Run E2E tests
+    needs: build-dists
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.9", "3.10", "3.11", "3.12", "3.13" ]
+        os: [ ubuntu-latest, macOS-latest, windows-latest ]
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch' }}
```
kyleknap, line 71: Could we wrap the `push` and main branch conditional in parenthesis? I just want to make sure those are always evaluated first and make that clear.

kyleknap (commented):
Looks good. Just had some more comments. I like the approach of just adding it to the `ci.yml`. I think that is going to make it managing the workflows easier.

.github/workflows/ci.yml
```diff
@@ -72,3 +72,42 @@ jobs:
...
+        with:
+          name: dists
+          path: dist/
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@39cd14951b08e74b54015e9e001cdefcf80e669f
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install -r requirements-dev.txt
```
kyleknap, line 101: Let's remove the line for installing from `requirements-dev.txt`. Specifically, for end to end tests, we want to install it as close as we can to what it would be like if installing directly from PyPI, which likely means unpinned dependencies. So, by removing this line, it will make sure we can install the library using the latest versions of our dependencies.

kyleknap (commented):
Looks good. Just had one more comment and we should be good to merge it.

PR #51, opened 2025-05-14
Added mypy

anjaliratnam-msft:
Added mypy and fixed some typing errors.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,6 @@
+# mypy.ini
+[mypy]
+check_untyped_defs = True
+disallow_untyped_defs = False
+ignore_missing_imports = True
+disable_error_code = union-attr, attr-defined, typeddict-unknown-key, no-any-return, return, assignment
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,3 +4,6 @@ check_untyped_defs = True
 disallow_untyped_defs = False
 ignore_missing_imports = True
 disable_error_code = union-attr, attr-defined, typeddict-unknown-key, no-any-return, return, assignment
+
+[mypy-tests.*]
+ignore_errors = True
--- a/src/azstoragetorch/_client.py
+++ b/src/azstoragetorch/_client.py
@@ -238,7 +238,7 @@ class AzStorageTorchBlobClient:
             max_in_flight_requests = self._get_max_in_flight_requests()
         self._max_in_flight_requests = max_in_flight_requests
         self._executor = executor
-        self._blob_properties = None
+        self._blob_properties: Optional[azure.storage.blob.BlobProperties] = None
 
     @property
     def url(self) -> str:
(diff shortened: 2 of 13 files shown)


mypy.ini
```diff
@@ -0,0 +1,9 @@
+# mypy.ini
+[mypy]
+check_untyped_defs = True
+disallow_untyped_defs = False
+ignore_missing_imports = True
+disable_error_code = union-attr, attr-defined, typeddict-unknown-key, no-any-return, return, assignment
+
+[mypy-tests.*]
+ignore_errors = True
```
kyleknap, line 9: Instead of ignoring the tests module, I'm thinking we just update the GitHub action to just run `mypy` on the `src` directory (e.g. `mypy src`). This is mainly so that if someone did want to `mypy` on the tests, it would not automatically ignore errors.

mypy.ini
```diff
@@ -0,0 +1,9 @@
+# mypy.ini
+[mypy]
+check_untyped_defs = True
+disallow_untyped_defs = False
+ignore_missing_imports = True
```
kyleknap, line 5: Do we need `ignore_missing_imports` option? Mainly asking this out of curiosity on what dependency is causing issues.

mypy.ini
```diff
@@ -0,0 +1,9 @@
...
+disable_error_code = union-attr, attr-defined, typeddict-unknown-key, no-any-return, return, assignment
```
kyleknap, line 6: We should avoid bulk disabling error codes. Mainly this will avoid scenarios where we introduce a new typing issue that is not caught because the error is globally disabled. Instead of using `disable_error_code`, I'm thinking we:

1. Try to fix the actual typing issue
2. If it is not possible or very impractical to fix the typing issue, we can add an inline ignore error code.

At some point, we should even consider adding the [`strict`]https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options) option, but I think that can be a follow up as that would likely require more changes and could be even more finicky to get working.

kyleknap (commented):
Looks good. Just had a few suggestions especially on the configuration of mypy

src/azstoragetorch/datasets.py
```diff
@@ -370,7 +370,7 @@ def __iter__(self) -> Iterator[_TransformOutputType_co]:
         worker_info = torch.utils.data.get_worker_info()
         for i, blob in enumerate(self._blobs):
             if self._should_yield_from_worker_shard(worker_info, i):
-                yield self._transform(blob)
+                yield self._transform(blob) # type: ignore[misc]
```
kyleknap, line 373: Can we remove the type ignores now that we are casting the default transform callable?

src/azstoragetorch/_client.py
```diff
@@ -396,7 +398,7 @@ def _download_from_unknown_blob_size(
             length = self._PARTITIONED_DOWNLOAD_THRESHOLD
         return self._download_with_retries(offset, length)
 
-    def _download_with_retries(self, pos: int, length: int) -> bytes:
+    def _download_with_retries(self, pos: int, length: int) -> bytes:   # type: ignore[return]
```
kyleknap, line 401: It would be interesting to see if we could remove this type ignore. Mainly if we had a bug where we actually were not returning `bytes` we would not catch that bug through type checking. Instead maybe we should consider raising an exception at the very end to say retries are exhausted (we would need to figure out the correct exception type we want)? I'm also open to other ideas but it seems like mypy is just not figuring out that the end of the method will never be reached.

kyleknap (commented):
Looks good. I just had some comments on the type updates. It would also be good to rebase or merge in the `ruff` format updates to get CI running for this PR and fix any merge conflicts.

PR #56, opened 2025-06-05
Upgrade dependencies

anjaliratnam-msft:
Upgraded the uv.lock file dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,12 +4,24 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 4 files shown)


kyleknap (commented):
In addition to updating the `uv.lock` let's:
* Increase the version ceiling of `azure-storage-blob` to the latest version of the SDK in the `pyproject.toml`
* Regenerate the `requirments-dev.txt` file using `uv` since that is what the CI jobs are using as of now for installing dependencies.

kyleknap (approved):
:ship:

PR #58, opened 2025-06-09
Run samples

anjaliratnam-msft:
Added bootstrap file to run the samples.


samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
+import os
```
kyleknap, line 1: Let's make sure to:
1. Add the license comment to the top of the script
2. Add a small docstring explaining this script is for development purposes only and serves as a way to automate running of samples.
3. Let's rename the script to `_run_sample.py`. Mainly using the verb `run` better indicates that we are running the script. Bootstrap indicates more that we are spinning up resources which we are not.

samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+                for item in var_value:
+                    print(item)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) > 1:
+        path = sys.argv[1]
+        run_sample(path)
+    else:
+        print("Please provide the path to the sample file")
```
kyleknap, line 42: Let's use [`argparse`](https://docs.python.org/3/library/argparse.html) for any command line parsing. It is going to be more robust than doing parsing ourselves. I used it in the scripts for the Jupyter notebook as an example: https://github.com/Azure/azure-storage-for-pytorch/blob/main/samples/intro_notebook/bootstrap.py.

samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+    new_url = rf"\1{STORAGE_ACCOUNT_NAME}\3{CONTAINER_NAME}\5"
+
+    return re.sub(pattern, new_url, content)
+
+
+def run_sample(path):
+    modified_content = modify_url(path)
+    if modified_content:
+        local_vars = {}
+        exec(modified_content, locals=local_vars)
```
kyleknap, line 29: Let's avoid calling `exec()`. Instead let's just use subprocesses such as [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) to invoke a python interpreter (e.g. by using `sys.executable` as the command to run). The main reasons that I'm suggesting this is running `exec` can lead to problematic side effects (e.g. mutating state in the current process) and should be avoided unless we really need to use it. In using subprocesses, the execution of the code is much more sandboxed to the child process and easier to manage.

samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+        print(
+            "Please set environment variables AZSTORAGETORCH_STORAGE_ACCOUNT_NAME and AZSTORAGETORCH_CONTAINER_NAME"
+        )
+        return
+
+    with open(path, "r") as f:
+        content = f.read()
+
+    pattern = r'(CONTAINER_URL\s*=\s*\(\s*"https://)([^.]+)(\.blob\.core\.windows\.net/)([^"]+)("\s*\))'
+    new_url = rf"\1{STORAGE_ACCOUNT_NAME}\3{CONTAINER_NAME}\5"
```
kyleknap, line 20: Instead of searching for the `CONTAINER_URL` value, can we instead search for `<my-storage-account-name>` and `<my-container-name>`? This will probably simplify the search pattern and more representative of what we are asking a customer to do (e.g. replace those placeholders)

samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+def run_sample(path):
+    modified_content = modify_url(path)
+    if modified_content:
+        local_vars = {}
+        exec(modified_content, locals=local_vars)
+
+        for var_name, var_value in local_vars.items():
+            if var_name == "map_dataset" or var_name == "iterable_dataset":
+                for item in var_value:
+                    print(item)
```
kyleknap, line 34: Then if we switch to using subprocesses, we should just make sure the subprocess exited with a return code of zero, which we should be able to turn on with `check=True`.

samples/_bootstrap.py
```diff
@@ -0,0 +1,42 @@
...
+        content = f.read()
+
+    pattern = r'(CONTAINER_URL\s*=\s*\(\s*"https://)([^.]+)(\.blob\.core\.windows\.net/)([^"]+)("\s*\))'
+    new_url = rf"\1{STORAGE_ACCOUNT_NAME}\3{CONTAINER_NAME}\5"
+
+    return re.sub(pattern, new_url, content)
+
+
+def run_sample(path):
+    modified_content = modify_url(path)
```
kyleknap, line 26: In order to use this with subprocesses, we can instead just write the modified sample content to a temporary file and use that file when executing it with our Python interpreter.

kyleknap (commented):
Looks like a good start. I just had some suggestions on how to approach it.

doc/sphinx/user-guide.rst
```diff
@@ -116,7 +116,7 @@ is performed using the `List Blobs API <List Blobs API_>`_. For example
         
         .. literalinclude:: ../../samples/map_dataset/dataset_from_container_url.py
             :lines: 9-
-
+
```
kyleknap, line 119: It looks like there is still some whitespace being added? To tell it is actually removed, I'd recommend running `git diff` against main to make sure this file does not change. In general, it would not hurt to run `git diff` prior to pushing changes to make sure all of the changes are accounted for or there is no unwanted changes.

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+            print(result.stdout)
+        except subprocess.CalledProcessError as e:
+            print(f"Error running sample: {e.stderr}")
+        finally:
+            os.remove(temp_path)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Run a sample script with a modified Azure Storage URL"
```
kyleknap, line 57: For the description, we should be able to just use the docstring of the file (e.g. `description=__doc__`) which will allow us to reuse the content. The main part is that I want to make sure that we include in the `--help` string that this for internal development purposes as well as in the docstring of the file

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+import re
+import subprocess
+import tempfile
+import argparse
+
+STORAGE_ACCOUNT_NAME = os.environ.get("AZSTORAGETORCH_STORAGE_ACCOUNT_NAME")
+CONTAINER_NAME = os.environ.get("AZSTORAGETORCH_CONTAINER_NAME")
+
+
+def modify_url(path):
```
kyleknap, line 20: Let's rename this function to be `update_placeholders` to better reflect that we are just updating places where there are placeholders, which technically could eventually be outside of a URL.

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+    if not STORAGE_ACCOUNT_NAME or not CONTAINER_NAME:
+        print(
```
kyleknap, line 22: Instead of printing out the error message, let's raise a `ValueError`. This is helpful because it will make sure we short circuit the rest of the script and return a non-zero error code.

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+
+    account_pattern = r"<my-storage-account-name>"
+    container_pattern = r"<my-container-name>"
+    modified_content = re.sub(account_pattern, STORAGE_ACCOUNT_NAME, content)
+    return re.sub(container_pattern, CONTAINER_NAME, modified_content)
+
+
+def run_sample(path):
+    modified_content = modify_url(path)
+    if modified_content:
```
kyleknap, line 38: Then if we raise an exception here, we should no longer need this `if` statement to execute the code if it changed.

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+        with tempfile.NamedTemporaryFile(
```
kyleknap, line 39: Instead of using a named temporary file, it might be worth using a `TemporaryDirectory`. For example, I used on in this [script](https://github.com/Azure/azure-storage-for-pytorch/blob/main/samples/intro_notebook/bootstrap.py#L55). The nice part of tempory directories is that we have more control over file creation (e.g. we can just use `open()` to create a file with a name we want).

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+            temp_path = temp_file.name
+        try:
+            result = subprocess.run(
+                [sys.executable, temp_path], capture_output=True, text=True, check=True
+            )
+            print(result.stdout)
+        except subprocess.CalledProcessError as e:
+            print(f"Error running sample: {e.stderr}")
+        finally:
+            os.remove(temp_path)
```
kyleknap, line 52: For any of the temporary files, we should not need to delete them ourselves as they are automatically deleted at exit on our behalf. So we should be able to remove this line.

samples/_run_sample.py
```diff
@@ -0,0 +1,61 @@
...
+        ) as temp_file:
+            temp_file.write(modified_content)
+            temp_path = temp_file.name
+        try:
+            result = subprocess.run(
+                [sys.executable, temp_path], capture_output=True, text=True, check=True
+            )
+            print(result.stdout)
+        except subprocess.CalledProcessError as e:
+            print(f"Error running sample: {e.stderr}")
```
kyleknap, line 50: We should be able to simplify this call to:
```python
subprocess.run([sys.executable, temp_path], check=True)
```
Mainly by default it will allow us to print output from standard output of the script and we do not necessarily need to catch `CalledProcessError` because stderr will be displayed as well automatically.

kyleknap (commented):
This is looking good. I just had some smaller suggestions.

PR #61, opened 2025-06-11
Added additional samples

anjaliratnam-msft:
Added additional samples from the user guide.

### DIFF
--- a/doc/sphinx/user-guide.rst
+++ b/doc/sphinx/user-guide.rst
@@ -222,7 +222,7 @@ returned across workers to avoid a :py:class:`~torch.utils.data.DataLoader` from
 duplicate samples from its workers
 
     .. literalinclude:: ../../samples/iterable_dataset/multiple_workers.py
-        :lines: 9-36
+        :lines: 9-
 
 
 .. _Azure subscription: https://azure.microsoft.com/free/
--- a/samples/iterable_dataset/multiple_workers.py
+++ b/samples/iterable_dataset/multiple_workers.py
@@ -9,13 +9,13 @@
 from azstoragetorch.datasets import IterableBlobDataset
 from torch.utils.data import DataLoader
 
+# Update URL with your own Azure Storage account and container name
+CONTAINER_URL = (
+    "https://<my-storage-account-name>.blob.core.windows.net/<my-container-name>"
+)
+
 
 def load_with_workers():
-    # Update URL with your own Azure Storage account and container name
-    CONTAINER_URL = (
-        "https://<my-storage-account-name>.blob.core.windows.net/<my-container-name>"
-    )
-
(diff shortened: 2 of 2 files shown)


samples/iterable_dataset/multiple_workers.py
```diff
@@ -0,0 +1,41 @@
...
+
+from azstoragetorch.datasets import IterableBlobDataset
+from torch.utils.data import DataLoader
+
+
+def load_with_workers():
+    # Update URL with your own Azure Storage account and container name
+    CONTAINER_URL = (
+        "https://<my-storage-account-name>.blob.core.windows.net/<my-container-name>"
+    )
```
kyleknap, line 17: For the `CONTAINER_URL` let's define it outside of the function as part of the top-level module. In general, when a variable is capital cased in Python, it serves as a constant and generally should be defined at the module level.

samples/iterable_dataset/multiple_workers.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # The number of samples returned from the dataset should be equal to the number of samples
+    # returned from the DataLoader. If the dataset did not handle sharding, the number of samples
+    # returned from the DataLoader would be ``num_workers`` times (i.e., four times) the number
+    # of samples in the dataset.
+    assert num_samples_from_dataset == num_samples_from_dataloader
+
+
+if __name__ == "__main__":
+    # Protects the entry point of the script and safely imports the module before calling the function
```
kyleknap, line 40: Let's reword this to make sure it is specifically due to the data loader using process-based workers. Maybe update it to be?
```
# Because the DataLoader uses processes for its workers, this if statement protects the script so that
# spawned processes can safely import the module without risk of calling `load_with_workers()` again.
```

doc/sphinx/user-guide.rst
```diff
@@ -251,32 +219,10 @@ When using a :py:class:`~azstoragetorch.datasets.IterableBlobDataset` and
...
-        num_samples_from_dataloader += len(batch["url"])
+duplicate samples from its workers
 
-    # The number of samples returned from the dataset should be equal to the number of samples
-    # returned from the DataLoader. If the dataset did not handle sharding, the number of samples
-    # returned from the DataLoader would be ``num_workers`` times (i.e., four times) the number
-    # of samples in the dataset.
-    assert num_samples_from_dataset == num_samples_from_dataloader
+    .. literalinclude:: ../../samples/iterable_dataset/multiple_workers.py
+        :lines: 9-36
```
kyleknap, line 225: I think it is fine to keep the lines at `9-` here. Mainly if someone was to copy and paste it and update the constant, it will work out of the box.

kyleknap (commented):
Looks good. I just had a few more suggestions. We should be good to merge after.

kyleknap (approved):
Looks good! :ship:

PR #94, opened 2026-04-29
compat with storage updates

l0lawrence:
tests failing bc the version testing is not released


src/azstoragetorch/_client.py
```diff
@@ -440,16 +438,22 @@ def _get_download_stream(self, pos: int, length: int) -> Iterator[bytes]:
...
             }
             if self._blob_properties is not None:
-                download_kwargs["modified_access_conditions"] = (
-                    azure.storage.blob._generated.models.ModifiedAccessConditions(
-                        if_match=self._blob_properties.etag
-                    )
-                )
+                download_kwargs["etag"] = self._blob_properties.etag
+                download_kwargs["match_condition"] = MatchConditions.IfNotModified
+            # Use cls callback to capture the pipeline response so we can read
```
l0lawrence, line 443: this could be setup differently I think to make it cleaner**

PR #95, opened 2026-09-16
Add azurite support for e2e testing

kyleknap:
To use it, specify `--azurite` when running the e2e suite e.g.,
```
py.test tests/e2e --azurite
```
The tests will automatically connect to the default settings of an azurite instance to run the end-to-end tests against. Also added it to CI to be able to run end to end tests for all PRs.

For azurite testing, an internal environment variable was also added to allow client request id's to not be echoed in the response because echoing is not supported as part of exception responses in Azurite.


tests/e2e/conftest.py
```diff
@@ -4,15 +4,59 @@
...
+    ResourceTypes,
+    Services,
+    generate_account_sas,
+)
+
 from tests.e2e.utils import random_resource_name
 
+_ALLOW_MISSING_CLIENT_REQUEST_ID_ENV_VAR = (
+    "_AZSTORAGETORCH_ALLOW_MISSING_CLIENT_REQUEST_ID"
+)
```
nateprewitt: Does it make sense to declare this in one place since it's also declared in the class in _client.py?

tests/e2e/conftest.py
```diff
@@ -21,11 +65,38 @@ def account_url():
...
+            permission=AccountSasPermissions(
+                read=True,
+                write=True,
+                delete=True,
+                list=True,
+                add=True,
+                create=True,
+                update=True,
+                process=True,
+            ),
```
nateprewitt: This is a little interesting that we're turning everything on for perms. Maybe there's not a straightforward way to toggle that less granularly.

tests/e2e/conftest.py
```diff
@@ -21,11 +65,38 @@ def account_url():
...
+
 @pytest.fixture(scope="package")
-def blob_service_client(account_url):
-    blob_service_client = BlobServiceClient(
-        account_url, credential=DefaultAzureCredential()
-    )
+def blob_service_client(account_url, credential):
+    if credential is None:
+        credential = DefaultAzureCredential()
+    blob_service_client = BlobServiceClient(account_url, credential=credential)
```
nateprewitt, line 90: Should this be `credential=None` if we're defaulting?

tests/unit/test_client.py
```diff
@@ -49,6 +49,9 @@
     azure.core.exceptions.DecodeError,
 ]
 PROCESS_CPU_COUNT_UNAVAILABLE = object()
+_ALLOW_MISSING_CLIENT_REQUEST_ID_ENV_VAR = (
+    "_AZSTORAGETORCH_ALLOW_MISSING_CLIENT_REQUEST_ID"
+)
```
nateprewitt: Same question here.

tests/e2e/conftest.py
kyleknap: I originally had it that way; the tests importing the constant from `_client.py`. Changed it to this more out of principal of not importing from private modules. Going though the diff right now, it does sort of stick out. I'll go ahead and just change it back from importing from `_client.py`

nateprewitt: It could either go there, or I don't know if you have any module used for constants/primitives for reuse. Given it appears in many places it may be helpful. Not pressing though.

tests/e2e/conftest.py
kyleknap: No. The Python SDK defaults `None` to anonymous access. So we need to set the default credentials here unfortunately. But for `azstoragetorch`, it's default of `None` results in default azure credentials being used.

Let me know if I interpreted this question/comment correctly.

tests/e2e/conftest.py
kyleknap: I could not find a straightforward way to "allow everything" in the SDK. I did decide to go ahead and remove some configuration though that was not actually needed. So that should simplify it a bit.

tests/e2e/conftest.py
nateprewitt: Sorry, I meant for the function signature. If we're always converting None to DefaultAzureCredential, should it be `def blob_service_client(account_url, credential=None):`?

nateprewitt (approved):
Left a response to one of the comments but nothing blocking.

tests/e2e/conftest.py
kyleknap: Ah since `credential` is a fixture, I don't think we want to set a default in the function signature because it seems like pytest won't do fixture injection if it sees a default and we need the `credential` fixture injected. I'm going to leave this as is.

PR #97, opened 2026-09-17
Add read-specific unit tests to e2e tests

kyleknap:
The unit tests cover more how blobio should be interacting with its internal client. The new e2e test cases give us more functional coverage on test cases if the backend for blobio was swapped out to better guarantee any backend changes are transparent.


nateprewitt (approved):
:shipit:

PR #98, opened 2026-09-18
Add write-specific unit tests to e2e tests

kyleknap:
Allows better coverage for end to end behavior if we change underlying implementation.


tests/e2e/test_write.py
```diff
@@ -14,34 +17,88 @@
...
-def downloaded_blob(container_client, blob_url):
-    blob_client = container_client.get_blob_client(blob=blob_url.split("/")[-1])
+@pytest.fixture
+def unauthorized_blob_io(blob_client):
+    # Assume the target container does not permit anonymous writes.
+    f = BlobIO(blob_client.url, "wb", credential=False)
+    yield f
+    # Suppress the repeated fatal error while cleaning up the file object.
+    with contextlib.suppress(FatalBlobIOWriteError):
+        f.close()
```
nateprewitt, line 40: Is this a behavior of the connector or is something else going wrong that's causing this to bubble up? What's making it repeat?

tests/e2e/test_write.py
```diff
@@ -14,34 +17,88 @@
...
+def _assert_staged_blocks(blob_client, expected_sizes):
+    try:
+        _, uncommitted_blocks = blob_client.get_block_list(
+            block_list_type="uncommitted"
+        )
+        assert sorted(block.size for block in uncommitted_blocks) == sorted(
+            expected_sizes
+        )
+    except ResourceNotFoundError:
+        assert expected_sizes == []
```
nateprewitt, line 79: nit; is hiding the exception not going to mask a missing resource with something that has no blocks staged?

nateprewitt (approved):
Two minor comments/clarifications. Otherwise looks good!

tests/e2e/test_write.py
kyleknap: This is the behavior of the connector; if any error is encountered, any subsequent mutating calls (e.g., `write()`, `flush()`, `close()`) will fail. This mainly due to writes being processed asynchronously and deciding to error on the side of simplicity and verbosity when it comes to bubbling errors up. I'm open to this changing though especially if writes no longer become asynchronous or downstream consumers find it too noisy.

tests/e2e/test_write.py
kyleknap: It should be fine. Given it requires an `expected_sizes` list, it should behave as expected:

* For a blob that does not exist nor has staged blocks, it should result in a `ResourceNotFoundError` and assert that `expected_sizes` is empty (i.e., no blocks have been staged). Mainly until a blob has at least one block staged, it will not exist as there is no initializer API (e.g., create multipart upload).
* For a blob that does exist (e.g., has already been committed) but has no staged blocks, it should result in no error being thrown an empty `uncommited_blocks` list being returned.

